PR #14, opened 2023-11-05
Add system tests and scale with larger queue size

samos123:


autoscaler.go
```diff
@@ -28,13 +28,17 @@ type Autoscaler struct {
...
 	for range time.Tick(a.Interval) {
 		log.Println("Calculating scales for all")
-		// TODO: Might need to account for queue size being 100
-		// probably can simply divide the average by queue size
 		for deploymentName, waitCount := range a.FIFO.WaitCounts() {
 			avg := a.getMovingAvgQueueSize(deploymentName)
 			avg.Next(float64(waitCount))
 			flt := avg.Calculate()
-			ceil := math.Ceil(flt)
+			// TODO fix this to use configurable concurrency setting that's supplied
```
nstogner, line 35: I think this TODO comment can be removed.

nstogner (approved):
Approved, I think one comment can be changed though

autoscaler.go
samos123: I rather keep it for myself since its not obvious what to change. I might forget in a few days when i try to make concurrency configurable

Issue #2, opened 2023-10-31
High Availability


nstogner:
Lingo currently assumes that it only has a single replica of itself running at a given point in time. If there were multiple replicas they might fight against eachother with setting the desired replica count for Deployments.

PR #16, opened 2023-11-07
HA Design

nstogner:
Design for #2


nstogner:
Thinking it might make sense to have the autoscaling leader aggregate queue sizes from the other lingo replicas by issuing http requests directly to those replicas instead of going through a ConfigMap. This would have less load on API Server. The tracking of per-Pod IPs is already built in due to how lingo keeps track of backends.

PR #20, opened 2023-11-11
Package refactor

### LINKED ISSUE #2: High Availability

nstogner:
Refactor into packages and prepare for HA.


cmd/lingo/main.go
```diff
@@ -72,64 +77,96 @@ func run() error {
...
 	}
 
-	fifo := queuemanager.NewFIFOQueueManager(concurrencyPerReplica)
+	clientset, err := kubernetes.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		return fmt.Errorf("clientset: %w", err)
+	}
+
+	const POD_NAME = "POD_NAME"
+	podName := os.Getenv(POD_NAME)
```
samos123, line 93: could we get this value automatically? or have a default value if POD_NAME is not set?

samos123: I think defaulting to hostname would be good enough, since that's essentially what you're doing in the YAML:
```
        - name: POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
```

nstogner: Good call, changed

deploy/proxy_deploy.yaml
```diff
@@ -16,5 +16,10 @@ spec:
     spec:
       serviceAccountName: proxy-controller
       containers:
-      - name: main
-        image: proxy-controller
+        - name: main
+          image: proxy-controller
+          env:
+            - name: NAMESPACE
```
samos123, line 22: does the proxy only work for 1 namespace?

samos123 (approved):

deploy/proxy_deploy.yaml
nstogner: Yes, you would need 1 proxy per namespace as currently setup. We can introduce a cluster-scoped proxy in the future.

PR #31, opened 2023-11-30
Load tests

nstogner:
The beginnings of load tests


skaffold.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: golang-proxy
 build:
   artifacts:
-    - image: lingo
+    - image: us-central1-docker.pkg.dev/substratus-dev/substratus/lingo
```
samos123, line 7: Any reason why you use GAR here? We have lingo published for free to ghcr.

nstogner: Not really, I wanted to be able to use skaffold from my local machine (I guess that doesnt matter here) and also didnt want to need to rely on other github actions running. Any thoughts?

samos123: In that case you can revert the change to the image and it should all just work

samos123: NVM I misunderstood how this worked outside of kind.

samos123 (approved):

PR #35, opened 2023-12-10
fix #34 error scale from 0 to 1

### LINKED ISSUE #34: error trying to scale after scaling up from 0 to 1
Note I believe the error seems to be benign but still we should handle this so the error isn't thrown.

Steps to reproduce:
1. Deploy lingo with these values:
```
concurrency: 50
scaleDownDelay: 300
```
2. Deploy mistral 7b instruct with vLLM deployment and min replicas 0 and max replicas = 10
3. Sent 500 concurrent requests

This message is seen non-stop (probably 500 times):
```
2023/12/10 07:12:44 Scaling model "mistral-7b-instruct-vllm": -1
2023/12/10 07:12:44 scale error: update scale (desired): Scale.autoscaling "mistral-7b-instruct-vllm" is invalid: spec.replicas: Invalid value: -1: must be greater than or equal to 0
2023/12/10 07:12:44 Entering queue a615bf79-11d2-401f-a97b-a8fac765b37b
2023/12/10 07:12:44 AtLeastOne()
2023/12/10 07:12:44 Scaling model "mistral-7b-instruct-vllm": -1
2023/12/10 07:12:44 scale error: update scale (desired): Scale.autoscaling "mistral-7b-instruct-vllm" is invalid: spec.replicas: Invalid value: -1: must be greater than or equal to 0
2023/12/10 07:12:44 Entering queue 743720cd-39ae-4317-991a-a4462cd2d121
```

Seems this is the code that's triggering the error:
```
func (s *scaler) AtLeastOne() {
	s.mtx.Lock()
	defer s.mtx.Unlock()
	log.Printf("AtLeastOne()")
	if err := s.scaleFunc(-1, true); err != nil {
		log.Printf("scale error: %v", err)
	}
}
```

This commit is probably the culprit: 6735594438f8016f1670567b71730d689c4446c5

samos123:
A benign error message is thrown when trying to atLeastOne gets called and there is already 1 or more replicas. This commit fixes that.

### DIFF
--- a/pkg/deployments/manager.go
+++ b/pkg/deployments/manager.go
@@ -111,7 +111,7 @@ func (r *Manager) getScaler(deploymentName string) *scaler {
 
 func (r *Manager) scaleFunc(ctx context.Context, deploymentName string) func(int32, bool) error {
 	return func(n int32, atLeastOne bool) error {
-		log.Printf("Scaling model %q: %v", deploymentName, n)
+		log.Printf("Scaling model %q: %v atLeastOne %v", deploymentName, n, atLeastOne)
 
 		req := types.NamespacedName{Namespace: r.Namespace, Name: deploymentName}
 		var d appsv1.Deployment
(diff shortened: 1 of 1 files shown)


nstogner:
Fix is included in #31

PR #37, opened 2023-12-15
Batch (with buckets) Design

nstogner:
Proposed design for a new Lingo batch API / control process.


nstogner:
Thinking it might be best to keep lingo lean and place this functionality elsewhere.

PR #38, opened 2023-12-15
Better concurrent request handling for model host address

alpe:
Like in #36 the reconcile may be affected by external requests. This refactoring helps by reducing lock conflicts 

* Handle request context timeout
* Optimise concurrent access in endpoints type by separating r/w lock and notification lock
* Added some tests and Go doc

I have also added a benchmark that shows that the new rwlock is ~30% faster than before on my box. But this is all within ns and does not really matter:
```
new: BenchmarkEndpointGroup-12    	 7667690	       154.6 ns/op
old: BenchmarkEndpointGroup-12    	 4968279	       234.2 ns/op
```
The key benefit of this PR is handling request timeout

### DIFF
--- a/pkg/endpoints/endpoints.go
+++ b/pkg/endpoints/endpoints.go
@@ -24,9 +24,8 @@ type endpointGroup struct {
 	ports     map[string]int32
 	endpoints map[string]endpoint
 
-	bmtx      sync.Mutex
-	bcast     chan struct{}   // closed when there's a broadcast
-	listeners []chan struct{} // keeps track of all listeners
+	bmtx  sync.RWMutex
+	bcast chan struct{} // closed when there's a broadcast
 }
 
 // getBestHost returns the best host for the given port name. It blocks until there are available endpoints
@@ -45,12 +44,10 @@ func (e *endpointGroup) getBestHost(ctx context.Context, portName string) (strin
 	// await endpoints exists
 	for len(e.endpoints) == 0 {
 		e.mtx.RUnlock()
-		_, err := execWithCtxAbort(ctx, func() any {
-			<-e.waitForEndpoints(ctx)
-			return nil
-		})
-		if err != nil {
-			return "", err
+		select {
+		case <-e.bcast:
+		case <-ctx.Done():
+			return "", ctx.Err()
 		}
 		e.mtx.RLock()
 	}
@@ -117,31 +114,10 @@ func (g *endpointGroup) setIPs(ips map[string]struct{}, ports map[string]int32) 
 	}
 }
 
-func (e *endpointGroup) waitForEndpoints(ctx context.Context) <-chan struct{} {
-	e.bmtx.Lock()
-	defer e.bmtx.Unlock()
-
-	ch := make(chan struct{})
-	e.listeners = append(e.listeners, ch)
-
-	go execWithCtxAbort(ctx, func() any {
-		<-e.bcast
-		close(ch)
-		return nil
-	})
-
-	return ch
-}
-
 func (g *endpointGroup) broadcastEndpoints() {
 	g.bmtx.Lock()
 	defer g.bmtx.Unlock()
 
 	close(g.bcast)
 	g.bcast = make(chan struct{})
-
-	for _, ch := range g.listeners {
-		close(ch)
-	}
-	clear(g.listeners)
 }
--- a/pkg/endpoints/endpoints.go
+++ b/pkg/endpoints/endpoints.go
@@ -1,6 +1,7 @@
 package endpoints
 
 import (
+	"context"
 	"fmt"
 	"sync"
 	"sync/atomic"
@@ -10,7 +11,7 @@ func newEndpointGroup() *endpointGroup {
 	e := &endpointGroup{}
 	e.ports = make(map[string]int32)
 	e.endpoints = make(map[string]endpoint)
-	e.active = sync.NewCond(&sync.Mutex{})
+	e.bcast = make(chan struct{})
 	return e
 }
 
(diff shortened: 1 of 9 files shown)


pkg/endpoints/manager.go
```diff
@@ -110,10 +110,31 @@ func (r *Manager) getEndpoints(service string) *endpointGroup {
...
+// It returns a slice of strings representing the hosts or an error if any context timeout occurred.
+func (r *Manager) GetAllHosts(ctx context.Context, service, portName string) ([]string, error) {
+	return execWithTimeout(ctx, func() []string { return r.getEndpoints(service).getAllHosts(portName) })
+}
+
+func execWithTimeout[T any](ctx context.Context, f func() T) (T, error) {
+	resultChan := make(chan T)
+	defer close(resultChan)
+	go func() {
+		resultChan <- f()
```
nstogner, line 131: Trying to grasp this function... Two concerns:

1. It appears that `resultChan` gets closed on context cancellation. If `f()` returns after cancellation wont it send a result on a closed channel and cause a panic?
2. If `f()` never returns, then wont we leak goroutines? I dont see what is causing `f()` to return on context cancellation.

pkg/endpoints/endpoints.go
```diff
@@ -19,19 +19,27 @@ type endpoint struct {
...
+//
+// It selects the host with the minimum in-flight requests among all the available endpoints.
+// The host is returned as a string in the format "IP:Port".
+//
+// Parameters:
+// - portName: The name of the port for which the best host needs to be determined.
+//
+// Returns:
+// - string: The best host with the minimum in-flight requests.
+func (e *endpointGroup) getBestHost(portName string) string {
```
nstogner, line 40: I think we are currently lacking the propagation of context cancellation down into this function where I think it needs to reach. Otherwise, this can hang forever (see comments in the timeout wrapper function). I am not sure you can accomplish this while utilizing the `sync.Cond` type. I think it might be best to heed the authors notes in the [sync.Cond docs](https://pkg.go.dev/sync#Cond) and find a solution that utilizes channels and respects context cancellation.

nstogner (changes requested):
Thanks for the PR! This is a tricky part of the codebase. I think we need a solution for waiting until an endpoint is available, but also respecting the context. I think there might be a leak in here as implemented - love to get your thoughts.

pkg/endpoints/manager.go
alpe: you are absolutely right with both of your points. I will refactor the code

alpe:
Thank you very much for the feedback! I applied some change but I think this needs some better testing before it can be merged. It is a much more complex beast than I thought initially

samos123:
Agree on the additional testing. Looks like the larger scale (300 concurrent requests) system test is catching some kind of issue: https://github.com/substratusai/lingo/actions/runs/7286882573/job/19856443938?pr=38#step:5:1098

I can help with additions to the system tests if you have specific scenarios that should be tested. The original system tests were a result of concurrent request handling being broken and needing to ensure scale up and scale down works as expected and request and responses are being returned for a realistic backend.

pkg/endpoints/manager.go
nstogner: I think this is still relevant. I would prefer to avoid this helper function and just remove the passed context to `GetAllEndpoints()` which appears to be the other caller of this helper func.

pkg/proxy/handler.go
```diff
@@ -56,7 +56,13 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	defer complete()
 
 	log.Println("Waiting for IPs", id)
-	host := h.Endpoints.GetHost(r.Context(), deploy, "http")
+	host, err := h.Endpoints.AwaitHostAddress(r.Context(), deploy, "http")
```
nstogner, line 59: It might be more robust to check this error instead of assuming is it a timeout error. In the future the error logic in the invoked function might be updated to return different error types but this call-site might not be reconsidered. Also, it is not always a timeout today: if the caller cancels the request the context will cancel (not technically a timeout).

pkg/endpoints/manager.go
alpe: Good idea. It is not blocking for any host 👍

pkg/endpoints/manager_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+func TestAwaitBestHost(t *testing.T) {
+	const myService = "myService"
+	const myPort = "myPort"
+
+	manager := &Manager{endpoints: make(map[string]*endpointGroup, 1)}
+	manager.getEndpoints(myService).
+		setIPs(map[string]struct{}{myService: {}}, map[string]int32{myPort: 1})
+
+	testCases := map[string]struct {
```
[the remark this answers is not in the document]
alpe: I agree with the error check a bit vague but IMHO it makes sense to have a spec for the methods that defines all cases. I find it more readable. 
But to be fair, I use table tests as my default structure for unit tests and may be biased. If this is very important for you, I can refactor. The error type is checked now

pkg/proxy/handler.go
```diff
@@ -56,7 +58,24 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		log.Printf("error while finding the host address %v", err)
+		switch {
+		case errors.Is(err, context.Canceled):
+			w.WriteHeader(http.StatusInternalServerError)
```
alpe, line 66: I did some research on the inet what http code makes sense but looks like this case is not handled explicit very often. Alternatively [`499`](https://http.dev/499) was suggested, which is not part of Go stdlib though

pkg/endpoints/manager_test.go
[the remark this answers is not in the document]
nstogner: Not a strong opinion, good with this.

PR #39, opened 2023-12-18
Add Prometheus metrics

### LINKED ISSUE #10: Publish metrics
Important metrics:

* Response times
* Current in-flight requests
* Current queue sizes
* Observed current replicas on backends
* Observed min/max configuration on backends

... etc

alpe:
Resolves #10 

Collects metrics from deployment, proxy and queue


nstogner:
Hey Alexander, welcome to the project and thanks for the PR! We needed metrics. Will review soon.

pkg/proxy/context.go
```diff
@@ -0,0 +1,25 @@
...
+type contextKey int
+
+const (
+	// position counter of the TX in the block
+	contextMmodelName contextKey = iota
+)
+
+// WithModelName sets the model name in the provided context and returns a new context with the updated value.
+// The model name can be retrieved using the contextMmodelName key.
+func WithModelName(ctx context.Context, name string) context.Context {
```
nstogner, line 15: In general, I try to avoid passing critical values through in context. I prefer to pass typed arguments to keep things clear and explicit. I would prefer we remove this part of the PR. I think the same functionality of publishing model-specific metrics (very useful) could be accomplished with helper functions called inside of the primary handler function.

pkg/deployments/manager.go
```diff
@@ -109,6 +109,16 @@ func (r *Manager) getScaler(deploymentName string) *scaler {
 	return b
 }
 
+func (r *Manager) iterateScalers(f func(deployment string, scaler *scaler) bool) {
```
nstogner, line 112: My preference would be for this function to be refactored to lock, iterate, unlock, and return a slice of metrics as opposed to being a generic mutex protected loop with a passed-in function. It made it hard for me to understand the expected lock-duration without visiting the call-site when written as-is.

cmd/lingo/main.go
```diff
@@ -112,6 +115,8 @@ func run() error {
 	le := leader.NewElection(clientset, hostname, namespace)
 
 	queueManager := queue.NewManager(concurrencyPerReplica)
+	metricsRegistry := prometheus.WrapRegistererWithPrefix("lingo_", metrics.Registry)
```
nstogner, line 118: Note: lingo actually already has a metrics exporter running. This is done in `ctrl.NewManager()` inside of `main()`, which exposes metrics about the kubernetes controller functionality. The actual package that starts the metrics server is [here](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.16.3/pkg/metrics/server). Here is a guide on how to expose additional metrics: https://book.kubebuilder.io/reference/metrics ... Thoughts on going that direction?

nstogner: PS: You can see these metrics by running lingo and curling:

```
curl localhost:8082/metrics | head
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 22737    0 22737# HELP certwatcher_read_certificate_errors_total Total number of certificate read errors
 # TYPE certwatcher_read_certificate_errors_total counter
 certwatcher_read_certificate_errors_total 0
 # HELP certwatcher_read_certificate_total Total number of certificate reads
# TYPE certwatcher_read_certificate_total counter
 certwatcher_read_certificate_total 0
0# HELP controller_runtime_active_workers Number of currently used workers per controller
 # TYPE controller_runtime_active_workers gauge
 controller_runtime_active_workers{controller="configmap"} 0
 controller_runtime_active_workers{controller="deployment"} 0
  0  2079k      0 --:--:-- --:--:-- --:--:-- 21.6M
```

nstogner (changes requested):
A much needed PR! Looking forward to getting this one merged after some changes.

cmd/lingo/main.go
alpe: Thanks for the link! I am using the global `metrics.Registry` as suggested in the book. The new metrics are returned by this endpoint already. It can be confusing but `WrapRegistererWithPrefix` is just some short cut to have the `lingo_` prefix attached to all custom metric names only. I don't have a strong opinion on the naming.

pkg/deployments/manager.go
alpe: Makes sense. Good feedback!

pkg/proxy/context.go
alpe: NP, I will refactor this.  👍 
My motivation to use the request context was to decouple the model name lookup from using the model name. This would provide an extension point via (http) middleware for people to integrate their own model parsing or custom logic when lingo is used as a library. But I realized that the same could be achieved by setting the `X-Model` header  anyway.

alpe:
I have applied the review feedback and added some tests. Thanks!

pkg/deployments/manager.go
```diff
@@ -109,6 +109,18 @@ func (r *Manager) getScaler(deploymentName string) *scaler {
 	return b
 }
 
+// getScalesSnapshot returns a snapshot of the stats for all scalers managed by the Manager.
+// The scales are returned as a map, where the keys are the model names
+func (r *Manager) getScalesSnapshot() map[string]scale {
+	r.scalersMtx.Lock()
```
alpe, line 115: fyi: a `sync.Map` would be a better fit instead of locking via `scalersMtx`

pkg/proxy/handler.go
```diff
@@ -24,15 +27,33 @@ type Handler struct {
...
 }
 
+func NewHandler(deployments *deployments.Manager, endpoints *endpoints.Manager, queues *queue.Manager) *Handler {
+	return &Handler{Deployments: deployments, Endpoints: endpoints, Queues: queues}
+}
+
 func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	var (
+		modelName          string
+		capturedStatusCode *int
```
nstogner, line 37: Why define `capturedStatusCode` as a pointer?

pkg/deployments/manager.go
nstogner: I sometimes wonder about this, but tend to agree with the comment in the docs on `sync.Map` about maintaining type safety:

> Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.

https://pkg.go.dev/sync#Map

nstogner (commented):
High quality stuff! Added a one question, then I think we are good to merge.

pkg/queue/manager.go
```diff
@@ -42,21 +53,23 @@ func (m *Manager) EnqueueAndWait(ctx context.Context, deploymentName, id string)
...
 func (m *Manager) getQueue(deploymentName string) *Queue {
-	m.mtx.Lock()
+	m.mtx.RLock()
 	q, ok := m.queues[deploymentName]
+	m.mtx.RUnlock()
 	if !ok {
-		q = New(m.conccurencyPerReplica)
-		go q.Start()
+		q = New(m.concurrencyPerReplica)
+		m.mtx.Lock()
```
alpe, line 69: Sorry, this is not safe without another read while write locked. I will undo

pkg/deployments/manager.go
alpe: Type safety is an argument. I will close the other PRs and may do another spike to propose a generic `SafeGetOrCreate(map, rwLock, key, factoryMethod)T` method. Write locks that can be triggered by external requests should be avoided

pkg/proxy/handler.go
alpe: This is pointing to the `statusCode` in `captureStatusResponseWriter`. But there is a simpler solution that may be more readable

PR #40, opened 2023-12-19
Add liveness, readiness probes to example deployment

### LINKED ISSUE #5: add /healthz endpoint that responds 200 OK

alpe:
Resolves #5 

This may not be the intended solution but tcp probes work well, too


alpe:
Not solving the readiness case

PR #43, opened 2023-12-20
Add readiness endpoint

### LINKED ISSUE #5: add /healthz endpoint that responds 200 OK

alpe:
Resolves #5


cmd/lingo/main.go
```diff
@@ -142,6 +142,12 @@ func run() error {
 	autoscaler.Endpoints = endpointManager
 	go autoscaler.Start()
 
+	readiness := deployments.NewReadiness(mgr.GetClient(), namespace, deploymentManager, 200*time.Millisecond)
```
alpe, line 145: The `namespace` is used to query model deployments as in the manager
`200ms` may be a bit optimistic

deploy/lingo_deploy.yaml
```diff
@@ -24,3 +24,14 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+          readinessProbe:
+            httpGet:
+              path: /readyz
+              port: 8081
```
alpe, line 30: The healthz endpoint was setup already in main. This adds port/path to make use of it. Not sure about the time  values here

deploy/lingo_deploy.yaml
```diff
@@ -24,3 +24,14 @@ spec:
...
+            initialDelaySeconds: 1
+            periodSeconds: 5
+          livenessProbe:
+            tcpSocket:
```
alpe, line 34: Tcp should be fine for liveness but if preferred, the health endpoint be configured to serve a liveness response, too
Again, not sure about the default time values here

pkg/deployments/manager.go
```diff
@@ -69,20 +69,13 @@ func (r *Manager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
...
-		}
-		models := strings.Split(modelCSV, ",")
-		if len(models) == 0 {
-			return ctrl.Result{}, nil
-		}
-		for _, model := range models {
-			r.setModelMapping(strings.TrimSpace(model), d.Name)
-		}
+	models := getModelsFromAnnotation(d.GetAnnotations())
+	if len(models) == 0 {
```
alpe, line 73: please note: this also ends the `ann==nil` path, unlike before

pkg/deployments/manager.go
```diff
@@ -98,6 +91,17 @@ func (r *Manager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
 	return ctrl.Result{}, nil
 }
 
+func getModelsFromAnnotation(ann map[string]string) []string {
```
alpe, line 94: extracted into a separate method to reuse in the readiness handler

pkg/deployments/readiness.go
```diff
@@ -0,0 +1,65 @@
...
+	deployments *Manager
+	interval    time.Duration // await deployments loaded on startup
+}
+
+func NewReadiness(k8sClient k8sClient, namespace string, deployments *Manager, interval time.Duration) *Readiness {
+	return &Readiness{k8sClient: k8sClient, namespace: namespace, deployments: deployments, interval: interval}
+}
+
+func (h *Readiness) StateLoaded(r *http.Request) error {
+	if h.ready.Load() {
```
alpe, line 31: my assumption here is, that a lingo instance is ready to serve requests once it has all the model deployments loaded. No need to check again.

pkg/deployments/readiness.go
```diff
@@ -0,0 +1,65 @@
...
+		return nil
+	}
+	return errors.New("not ready")
+}
+
+func (h *Readiness) WatchModelBackendsLoaded() {
```
alpe, line 37: This is called by main once and loops until all model backend deployments are in the deployment manager loaded. This may not work well when there are a lot of concurrent create/ deletes of deployments

pkg/deployments/readiness.go
```diff
@@ -0,0 +1,65 @@
...
+		var sliceList appsv1.DeploymentList
+		if err := h.k8sClient.List(context.TODO(), &sliceList, client.InNamespace(h.namespace)); err != nil {
+			log.Printf("failed to list deployments: %v", err)
+			continue
+		}
+		found := true
+	outer:
+		for _, v := range sliceList.Items {
+			models := getModelsFromAnnotation(v.Annotations)
+			if len(models) == 0 {
```
alpe, line 48: ignores all deployments without lingo model annotation

pkg/deployments/readiness.go
```diff
@@ -0,0 +1,65 @@
...
+				if _, ok := h.deployments.ResolveDeployment(m); !ok {
+					found = false
+					break outer
+				}
+			}
+		}
+
+		if found {
+			h.ready.Store(true)
+			return
```
alpe, line 62: end loop when all models found in state

nstogner:
Much needed! Will give this a thorough review soon.

nstogner (commented):
This is a very important feature to have!

I have a question on whether the client returned from `mgr` is ready to use before the call to `mgr.Start()`... It is a cache-backed client and I am not sure the cache will be synced when the readiness check calls it (perhaps leading to a `.List()` that returns fewer than actual items - resulting in a false-positive readiness check).

By default, when the controller starts up, it will automatically reconcile all Deployments... as far as I could tell there is no way to check when that initial run has completed, so I think it makes sense to do what you are doing here. However, I wonder if a cleaner way of accomplishing this would be to run a one-time `List()` (after cache is synced - or using a non-cache client) and iterate through and call `Reconcile()` (i.e. a forced full-reconciliation), then register an always-200 readiness handler.

PS: I don't know if there is any value in using this controller-runtime pkg to do the health checks or not, but might be worth a look: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/healthz#Handler

alpe:
Good questions!
I did some manual tests before but the caching may be outdated indeed. A single `List()` seems more effective with the `Reconcile()` called. 

I had looked at  https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/healthz#Handler before but this is just a generic framework to customize the healthz server handlers. Most examples point to the Ping handler as generic `200` response.

cmd/lingo/main.go
```diff
@@ -188,6 +193,11 @@ func run() error {
 		setupLog.Info("manager stopped")
 	}()
 
+	_ = mgr.GetCache().WaitForCacheSync(ctx)
```
alpe, line 196: The mgr client will fail when the cache is not synced. Alternatively a client without cache could be passed.

cmd/lingo/main.go
```diff
@@ -103,10 +105,15 @@ func run() error {
 		return fmt.Errorf("starting manager: %w", err)
 	}
 
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
```
alpe, line 108: This is the always  `200` handler

pkg/deployments/manager.go
```diff
@@ -174,6 +182,25 @@ func (r *Manager) ResolveDeployment(model string) (string, bool) {
...
 
+type k8sClient interface {
+	List(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error
+}
+
+// Bootstrap initializes the Manager by retrieving a list of deployments from the k8s cluster and adding them to the Manager's internal state.
+// The k8s client passed should either be non caching or have the cache synced first.
+func (r *Manager) Bootstrap(ctx context.Context, k8sClient k8sClient) error {
+	var sliceList appsv1.DeploymentList
+	if err := k8sClient.List(context.TODO(), &sliceList, client.InNamespace(r.Namespace)); err != nil {
```
alpe, line 193: This queries all deployments once.

pkg/deployments/manager.go
```diff
@@ -69,33 +69,41 @@ func (r *Manager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
...
-		}
-		for _, model := range models {
-			r.setModelMapping(strings.TrimSpace(model), d.Name)
-		}
-	}
+	return ctrl.Result{}, r.addDeployment(ctx, d)
+}
 
+func (r *Manager) addDeployment(ctx context.Context, d appsv1.Deployment) error {
+	models := getModelsFromAnnotation(d.GetAnnotations())
```
alpe, line 76: This method is not called twice on startup for each deployment. The code be optimized to exit early for know deployments

cmd/lingo/main.go
nstogner: We should add an error check here right?

pkg/deployments/manager.go
```diff
@@ -174,6 +182,25 @@ func (r *Manager) ResolveDeployment(model string) (string, bool) {
 	return deploy, ok
 }
 
+type k8sClient interface {
+	List(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error
+}
+
+// Bootstrap initializes the Manager by retrieving a list of deployments from the k8s cluster and adding them to the Manager's internal state.
+// The k8s client passed should either be non caching or have the cache synced first.
+func (r *Manager) Bootstrap(ctx context.Context, k8sClient k8sClient) error {
```
nstogner, line 191: `Manager` has a `Client` field. Can we use that instead of accepting a client arg?

pkg/deployments/manager.go
nstogner, line 193: This should take the passed `ctx`.

cmd/lingo/main.go
nstogner: Does this start serving after `mgr.Start()`? If so, I think we need to move `.Bootstrap()` above `mgr.Start()`.

alpe: This is a bit of a chicken/ egg problem. The `mgr.Start()` call also starts the cache syncs that we need for the `bootstrap` call. So with this implementation, the container may be "ready" but not `alive`, yet.
This can be solved with a custom handler that checks bootstrap was called once.
Good 👁️

PR #44, opened 2023-12-20
Handle model undeployment

alpe:
🚧 spike

Simple version to fail the request fast when a model backed was undeployed while queued. 

Not handled is the case when a model was removed from the deployment annotation

### DIFF
--- a/pkg/deployments/manager.go
+++ b/pkg/deployments/manager.go
@@ -11,8 +11,8 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
-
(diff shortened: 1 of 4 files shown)


pkg/deployments/manager.go
```diff
@@ -65,7 +65,11 @@ func (r *Manager) SetDesiredScale(deploymentName string, n int32) {
 
 func (r *Manager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var d appsv1.Deployment
-	if err := r.Get(ctx, req.NamespacedName, &d); err != nil {
+	switch err := r.Get(ctx, req.NamespacedName, &d); {
+	case apierrors.IsNotFound(err):
```
alpe, line 69: Reconcile is also called when a deployment is deleted. The name is unknown to the k8s api so we know it is deleted

nstogner (commented):
Good stuff, added some minor comments

PR #51, opened 2024-01-09
Spike: Add optional retry middleware

### LINKED ISSUE #48: Lingo should retry on proxy failure
There are many reasons a backend model server can fail to serve a request. If lingo adds retries on failure it could improve the overall reliability of the system.

alpe:
Resolves #48 

the implementation is based on 2 elements:
* server middleware to retry a failed request with http status code 502, 503, 504

not considered is a priority queue for retries

### DIFF
--- a/cmd/lingo/main.go
+++ b/cmd/lingo/main.go
@@ -119,11 +119,10 @@ func run() error {
 	metricsRegistry := prometheus.WrapRegistererWithPrefix("lingo_", metrics.Registry)
 	queue.NewMetricsCollector(queueManager).MustRegister(metricsRegistry)
 
-	endpointManager, err := endpoints.NewManager(mgr)
+	endpointManager, err := endpoints.NewManager(mgr, queueManager.UpdateQueueSizeForReplicas)
 	if err != nil {
 		return fmt.Errorf("setting up endpoint manager: %w", err)
 	}
-	endpointManager.EndpointSizeCallback = queueManager.UpdateQueueSizeForReplicas
 	// The autoscaling leader will scrape other lingo instances.
 	// Exclude this instance from being scraped by itself.
 	endpointManager.ExcludePods[hostname] = struct{}{}
--- a/cmd/lingo/main.go
+++ b/cmd/lingo/main.go
@@ -113,6 +113,7 @@ func run() error {
 	if err != nil {
 		return fmt.Errorf("getting hostname: %v", err)
 	}
+
 	le := leader.NewElection(clientset, hostname, namespace)
 
 	queueManager := queue.NewManager(concurrencyPerReplica)
@@ -152,7 +153,8 @@ func run() error {
 	go autoscaler.Start()
 
 	proxy.MustRegister(metricsRegistry)
-	proxyHandler := proxy.NewHandler(deploymentManager, endpointManager, queueManager)
+	var proxyHandler http.Handler = proxy.NewHandler(deploymentManager, endpointManager, queueManager)
+	proxyHandler = proxy.NewRetryMiddleware(3, proxyHandler)
 	proxyServer := &http.Server{Addr: ":8080", Handler: proxyHandler}
 
 	statsHandler := &stats.Handler{
--- a/cmd/lingo/main.go
+++ b/cmd/lingo/main.go
@@ -71,11 +71,13 @@ func run() error {
 	var metricsAddr string
 	var probeAddr string
 	var concurrencyPerReplica int
+	var maxRetriesOnErr int
 
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8082", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.IntVar(&concurrencyPerReplica, "concurrency", concurrency, "the number of simultaneous requests that can be processed by each replica")
 	flag.IntVar(&scaleDownDelay, "scale-down-delay", scaleDownDelay, "seconds to wait before scaling down")
+	flag.IntVar(&maxRetriesOnErr, "max-retries", 0, "max number of retries on a http error code: 502,503,504")
 	opts := zap.Options{
 		Development: true,
 	}
@@ -154,7 +156,7 @@ func run() error {
 
 	proxy.MustRegister(metricsRegistry)
 	var proxyHandler http.Handler = proxy.NewHandler(deploymentManager, endpointManager, queueManager)
-	proxyHandler = proxy.NewRetryMiddleware(3, proxyHandler)
+	proxyHandler = proxy.NewRetryMiddleware(maxRetriesOnErr, proxyHandler)
 	proxyServer := &http.Server{Addr: ":8080", Handler: proxyHandler}
 
 	statsHandler := &stats.Handler{
--- a/pkg/endpoints/endpoints.go
+++ b/pkg/endpoints/endpoints.go
@@ -2,7 +2,9 @@ package endpoints
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"sync/atomic"
 )
@@ -16,7 +18,8 @@ func newEndpointGroup() *endpointGroup {
 }
 
 type endpoint struct {
-	inFlight *atomic.Int64
+	inFlight   *atomic.Int64
+	terminated chan struct{}
 }
 
 type endpointGroup struct {
@@ -104,12 +107,13 @@ func (g *endpointGroup) setIPs(ips map[string]struct{}, ports map[string]int32) 
 	g.ports = ports
 	for ip := range ips {
 		if _, ok := g.endpoints[ip]; !ok {
-			g.endpoints[ip] = endpoint{inFlight: &atomic.Int64{}}
+			g.endpoints[ip] = endpoint{inFlight: &atomic.Int64{}, terminated: make(chan struct{})}
 		}
 	}
-	for ip := range g.endpoints {
+	for ip, endpoint := range g.endpoints {
 		if _, ok := ips[ip]; !ok {
 			delete(g.endpoints, ip)
+			close(endpoint.terminated)
 		}
 	}
 	g.mtx.Unlock()
@@ -126,4 +130,30 @@ func (g *endpointGroup) broadcastEndpoints() {
 
 	close(g.bcast)
 	g.bcast = make(chan struct{})
+}
+
+func (e *endpointGroup) AddInflight(addr string, cancelRequest context.CancelFunc) (func(), error) {
+	tokens := strings.Split(addr, ":")
+	if len(tokens) != 2 {
+		return nil, errors.New("unsupported address format")
+	}
+	e.mtx.RLock()
+	endpoint, ok := e.endpoints[tokens[0]]
+	e.mtx.RUnlock()
+	if !ok {
+		return nil, errors.New("unsupported endpoint address")
+	}
+	endpoint.inFlight.Add(1)
+	done := make(chan struct{})
+	go func() {
+		select {
+		case <-done:
+		case <-endpoint.terminated:
+			cancelRequest()
+		}
+	}()
+	return func() {
+		close(done)
+		endpoint.inFlight.Add(-1)
+	}, nil
 }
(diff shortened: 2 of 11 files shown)


pkg/proxy/handler.go
```diff
@@ -83,13 +87,36 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	log.Println("Waiting for IPs", id)
-	host := h.Endpoints.GetHost(r.Context(), deploy, "http")
+	host, err := h.Endpoints.AwaitHostAddress(r.Context(), deploy, "http")
+	if err != nil {
+		log.Printf("timeout finding the host address %v", err)
+		w.WriteHeader(http.StatusRequestTimeout)
```
samos123, line 93: Did I understand it correctly that right now there is no timeout configured? Only the test cases do set a timeout?

samos123 (commented):
Nice! This looks good to me. @nstogner is the expert in this area so will wait for his approval.

pkg/proxy/handler.go
```diff
@@ -83,13 +87,36 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		return func(w http.ResponseWriter, r *http.Request) {
+			newCtx, done, err := endpoints.RegisterInFlight(r.Context(), deploy, host)
+			if err != nil {
+				log.Printf("error registering in-flight request: %v", err)
+				w.WriteHeader(http.StatusInternalServerError)
+				return
+			}
+			defer done()
+
+			other.ServeHTTP(w, r.Clone(newCtx))
```
samos123, line 117: I don't understand where the retry for 503 is happening in the middleware. Could you explain how the retry mechanism works?

I was looking for some part of the code that would catch an error response (e.g. all 5xx errors) and then send a new request to a new endpoint.

alpe: The retry is not implemented yet. This is the first part where an inflight proxy request is canceled. I will push some changes today to add retry middleware

pkg/proxy/handler.go
alpe: There is no explicit timeout defined. But it can make sense to define this at a higher level for all requests.

pkg/proxy/middleware.go
```diff
@@ -0,0 +1,77 @@
...
+	"log"
+	"math/rand"
+	"net/http"
+	"time"
+)
+
+var _ http.Handler = &RetryMiddleware{}
+
+type RetryMiddleware struct {
+	other      http.Handler
```
samos123, line 14: `other` was a bit confusing to me at first. I don't have a suggestion either so feel free to ignore this comment.

pkg/proxy/handler.go
samos123: agreed, let's capture this in a separate issue, I can file it. I think a default timeout of 30 minutes would make sense with an option to set it lower

pkg/proxy/middleware.go
alpe: Good 👁️ . It is the next handler in the chain. I will give it a better name

pkg/endpoints/endpoints.go
```diff
@@ -70,28 +97,63 @@ func (e *endpointGroup) getPort(portName string) int32 {
...
+
+func (g *endpointGroup) broadcastEndpoints() {
+	g.bmtx.Lock()
+	defer g.bmtx.Unlock()
+
+	close(g.bcast)
+	g.bcast = make(chan struct{})
+}
+
+func (e *endpointGroup) AddInflight(addr string, cancelRequest context.CancelFunc) (func(), error) {
```
alpe, line 135: Not sure if it makes sense to actively cancel any inflight request here. We could also rely on the k8s internals to handle it.

pkg/proxy/middleware.go
```diff
@@ -0,0 +1,77 @@
...
+			header: make(http.Header),
+			body:   bytes.NewBuffer([]byte{}),
+		}
+		// call next handler in chain
+		r.nextHandler.ServeHTTP(capturedResp, request.Clone(request.Context()))
+
+		if i == r.MaxRetries || // max retries reached
+			request.Context().Err() != nil || // abort early on timeout, context cancel
+			capturedResp.status != http.StatusBadGateway &&
+				capturedResp.status != http.StatusServiceUnavailable {
```
alpe, line 43: Should we include `StatusGatewayTimeout`?

pkg/endpoints/endpoints.go
nstogner: I would avoid actively cancelled to avoid complexity

pkg/proxy/middleware.go
nstogner: Yes, I think `408`, `502`, `503`, `504` make sense. Perhaps a helper func `isRetryableStatusCode(int) bool`?

pkg/proxy/middleware.go
```diff
@@ -0,0 +1,77 @@
...
+		rSource:     rand.New(rand.NewSource(time.Now().UnixNano())),
+	}
+}
+
+func (r RetryMiddleware) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	var capturedResp *responseBuffer
+	for i := 0; ; i++ {
+		capturedResp = &responseBuffer{
+			header: make(http.Header),
+			body:   bytes.NewBuffer([]byte{}),
```
nstogner, line 35: As we spoke about, I think it would be good to know whether we are adding to in-memory buffering or not by doing this. I think it is still worth having this feature even if we are making memory requirements worse. This makes me think that we might actually want to make retrying configurable:

* enabled/disabled
* list of status codes to retry? maybe this configurable item would be excessive?

pkg/proxy/handler.go
samos123: I filed this issue: https://github.com/substratusai/lingo/issues/58

cmd/lingo/main.go
```diff
@@ -153,7 +153,8 @@ func run() error {
 	go autoscaler.Start()
 
 	proxy.MustRegister(metricsRegistry)
-	proxyHandler := proxy.NewHandler(deploymentManager, endpointManager, queueManager)
+	var proxyHandler http.Handler = proxy.NewHandler(deploymentManager, endpointManager, queueManager)
+	proxyHandler = proxy.NewRetryMiddleware(3, proxyHandler)
```
alpe, line 157: todo: hard coded 3 retries but it would be better to have a start param to make this configurable. `0` would disable the middleware

pkg/proxy/middleware.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		totalRetries.Inc()
+		// Exponential backoff
+		jitter := time.Duration(r.rSource.Intn(50))
+		time.Sleep(time.Millisecond*time.Duration(1<<uint(i)) + jitter)
+	}
+}
+
+func isRetryableStatusCode(status int) bool {
+	return status == http.StatusBadGateway ||
```
alpe, line 53: the values are still hardcoded here. I will make them configurable. 
Not sure about [`408`](https://http.dev/408). When the client is too slow, would a retry help?

nstogner: I am good with leaving off 408

cmd/lingo/main.go
nstogner: Sounds like a good configurable option. I am fine with hardcoding everything in this PR and adding another PR for configuration.

pkg/proxy/middleware.go
```diff
@@ -0,0 +1,255 @@
...
+		headerBuf:             make(http.Header),
+		discardErrResp:        discardErrResp, // abort early on timeout, context cancel
+	}
+	if _, ok := writer.(io.ReaderFrom); ok {
+		return &xResponseWriterDelegator{responseWriterDelegator: d}
+	}
+	return d
+}
+
+func (r *responseWriterDelegator) discardedResponse() bool {
```
nstogner, line 118: Seems like this method could go away in favor of just checking `r.discardErrResp`?

pkg/proxy/middleware.go
```diff
@@ -0,0 +1,255 @@
...
+
+// newResponseWriterDelegator constructor
+func newResponseWriterDelegator(writer http.ResponseWriter, isRetryableStatusCode func(status int) bool, discardErrResp bool) xResponseWriter {
+	d := &responseWriterDelegator{
+		isRetryableStatusCode: isRetryableStatusCode,
+		ResponseWriter:        writer,
+		headerBuf:             make(http.Header),
+		discardErrResp:        discardErrResp, // abort early on timeout, context cancel
+	}
+	if _, ok := writer.(io.ReaderFrom); ok {
```
nstogner, line 112: In Lingo when would the writer be an `io.ReaderFrom` and when would it not be?

nstogner (commented):
I didnt get enough time to fully go through this PR and understand it. I added a few questions will revisit.

pkg/endpoints/endpoints.go
```diff
@@ -16,7 +18,8 @@ func newEndpointGroup() *endpointGroup {
 }
 
 type endpoint struct {
-	inFlight *atomic.Int64
+	inFlight   *atomic.Int64
+	terminated chan struct{}
```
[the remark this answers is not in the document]
alpe: argh, this is a leftover from terminating in flight requests when the IP is removed. I will remove this. Thanks!

pkg/proxy/handler.go
```diff
@@ -117,10 +129,16 @@ func parseModel(r *http.Request) (string, *http.Request, error) {
 	if model := r.Header.Get("X-Model"); model != "" {
 		return model, r, nil
 	}
-	// parse request body for model name, ignore errors
-	body, err := io.ReadAll(r.Body)
-	if err != nil {
-		return "", r, nil
+	var body []byte
+	if mb, ok := r.Body.(*lazyBodyCapturer); ok && mb.capturedBody != nil {
```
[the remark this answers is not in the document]
alpe: With this change, I am re-using the buffer of the retry middleware so that the body must not be re-read (and buffered) again. It is an optimization

pkg/proxy/middleware.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	statusCodeIndex := make(map[int]struct{}, len(optRetryStatusCodes))
+	for _, c := range optRetryStatusCodes {
+		statusCodeIndex[c] = struct{}{}
+	}
+	return &RetryMiddleware{
+		nextHandler:      other,
+		maxRetries:       maxRetries,
+		retryStatusCodes: statusCodeIndex,
+		rSource:          rand.New(rand.NewSource(time.Now().UnixNano())),
```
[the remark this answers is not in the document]
alpe: You are right. The global rand is good enough. 👍 
I had run into some deprecated methods and went into some rabbit hole.

pkg/proxy/middleware.go
alpe: The `response` type in the Go stdlib implements the interface. So should do all decorators to make use of internal or future optimizations. it is used in `io.Copy` for example. I saw a presentation once but I don't have the link anymore. I also double checked with the prometheus [decorator](https://github.com/prometheus/client_golang/blob/main/prometheus/promhttp/delegator.go#L105) to ensure it is still a good practice

Same for `WriteTo` on the `lazyBodyCapturerWriteTo`

pkg/proxy/middleware.go
alpe: I need this to satisfy the response type in `newResponseWriterDelegator() xResponseWriter`. The source writer either implements `ReaderFrom` or not. So does the returned implementation.

pkg/proxy/middleware.go
```diff
@@ -0,0 +1,253 @@
...
+	retryStatusCodes map[int]struct{}
+}
+
+// NewRetryMiddleware creates a new HTTP middleware that adds retry functionality.
+// It takes the maximum number of retries, the next handler in the middleware chain,
+// and an optional list of retryable status codes.
+// If the maximum number of retries is 0, it returns the next handler without adding any retries.
+// If the list of retryable status codes is empty, it uses a default set of status codes (http.StatusBadGateway, http.StatusServiceUnavailable, http.StatusGatewayTimeout).
+// The function creates a RetryMiddleware struct with the given parameters and returns it as an http.Handler.
+func NewRetryMiddleware(maxRetries int, other http.Handler, optRetryStatusCodes ...int) http.Handler {
```
alpe, line 26: We need to limit the max bytes read.  See https://cwe.mitre.org/data/definitions/400.html example 6

nstogner: I can see this being useful.

alpe:
Closing this now in favour of #61

PR #52, opened 2024-01-10
Spike: Configurable concurrency

alpe:
🚧 Spike 

After substratusai#43

Resolves substratusai#12


pkg/queue/queue.go
```diff
@@ -9,21 +9,23 @@ import (
...
 	}
 }
 
 // Queue is a thread-safe FIFO queue that will limit the number of concurrent
 // requests that can be in progress at any given time.
 type Queue struct {
-	// concurrency is the max number of in progress items.
-	concurrency    int
 	concurrencyMtx sync.RWMutex
+	// concurrency is the max number of in progress items.
```
nstogner, line 26: What is the reason for having both `concurrency` and `concurrencyPerReplica`?

pkg/queue/queue.go
```diff
@@ -9,21 +9,23 @@ import (
...
+	concurrency           int
+	concurrencyPerReplica uint32
```
samos123, line 28: please add a comment for `concurrencyPerReplica` so people reading directly understand what it does.

e.g. add 
```
// concurrencyPerReplica is the max number of in progress items per replica.
```

pkg/queue/queue.go
```diff
@@ -166,19 +168,34 @@ func (q *Queue) GetConcurrency() int {
...
+// setConcurrencyByPerReplica updates the total concurrency and per replica value
+func (q *Queue) setConcurrencyByPerReplica(newPerReplica uint32) {
+	q.concurrencyMtx.Lock()
+	defer q.concurrencyMtx.Unlock()
+
+	if q.concurrencyPerReplica == newPerReplica {
+		return
+	}
+
+	replicas := q.concurrency / int(q.concurrencyPerReplica)
```
samos123, line 186: Shouldn't we get number of replicas from the system? The number of replicas fluctuates frequently and I think replicas counted this way would often not match the replicas in the actual system.

Maybe it should take a 2nd parameter: `replicas int`, because the deploy manager has the actual current replicas that kubernetes reports.

pkg/queue/queue.go
samos123, line 28: I would prefer to use `int` everywhere to keep it the same. No strong preference though.

samos123 (changes requested):

pkg/queue/manager.go
```diff
@@ -65,9 +69,23 @@ func (m *Manager) getQueue(deploymentName string) *Queue {
...
 }
+
+func (r *Manager) GetCurrencyPerReplica(deployment string) uint32 {
+	return r.getQueue(deployment).concurrencyPerReplica
+}
+
+// SetCurrencyPerReplica updates the concurrency value per replica for a specific deployment.
+// If the provided value is 0, it sets the concurrency value to the default value.
+// This function updates the concurrency value directly on the queue associated with the deployment.
+func (r *Manager) SetCurrencyPerReplica(deploymentName string, newPerReplica uint32) {
```
samos123, line 86: typo? should this be `SetConcurrencyPerReplica` instead of `SetCurrencyPerReplica`?

pkg/queue/manager.go
```diff
@@ -65,9 +69,23 @@ func (m *Manager) getQueue(deploymentName string) *Queue {
...
 	if !ok {
-		q = New(m.concurrencyPerReplica)
+		q = New(int(m.defaultConcurrencyPerReplica), m.defaultConcurrencyPerReplica)
 		m.queues[deploymentName] = q
 		go q.Start()
 	}
 	return q
 }
+
+func (r *Manager) GetCurrencyPerReplica(deployment string) uint32 {
```
samos123, line 79: typo? should this be `GetConcurrencyPerReplica` instead of `GetCurrencyPerReplica`?

pkg/queue/queue.go
samos123: or maybe we should change to `uint` everywhere? Whatever we do, ideally we do it consistently unless there is a strong reason not to.

pkg/queue/queue.go
alpe: @samos123  is correct. Let's have better names. I have updated the doc already:
```
// // concurrency is the max number of in progress items on all replicas. Same as `number_replicas * concurrencyPerReplica`
// concurrencyPerReplica is the max number of in progress items that a single replica instance can handle.
```

pkg/queue/queue.go
alpe: +1 for consistency. I will use `int` here and we change it in a new PR

samos123:

PR #54, opened 2024-01-11
General housekeeping

nstogner:
* Update readme
* Change test names / directories
* Update development.md guide

NOTE: No actual changes to testing code were made. Changes to workflow files are mostly for readability of files and also of the checks as displayed on PRs.


.github/workflows/build-push.yml
```diff
@@ -1,4 +1,4 @@
-name: Create and publish a Docker image
+name: Build and Push
```
samos123, line 1: Build and Push Container Image?

.github/workflows/build-push.yml
```diff
@@ -8,44 +8,59 @@ on:
...
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: substratusai/lingo
 
-# There is a single job in this workflow. It's configured to run on the latest available version of Ubuntu.
+# There is a single job in this workflow.
+# It's configured to run on the latest available version of Ubuntu.
 jobs:
-  build-and-push-image:
+  container:
```
samos123, line 24: container as job name seems less descriptive than `build-and-push-image`?

.github/workflows/load-tests.yml
```diff
@@ -4,34 +4,43 @@ run-name: Load tests by @${{ github.actor }}
 on: workflow_dispatch
 
 jobs:
-  perf-tests:
+  k6:
```
samos123, line 7: I probably prefer perf-tests or load-tests. I am not familiar with k6 but if it says load-test or perf-test I would have a general idea of what's supposed to happen in the job.

README.md
```diff
@@ -1,36 +1,41 @@
-# Lingo - K8s LLM Proxy + Scaler
+# Lingo - The lightweight model proxy
```
samos123, line 1: I do think we should add K8s in the title so it's clear that this only works on K8s and is very focused towards K8s.

What about one of these options?
- The lightweight ML Model proxy and autoscaler for K8s
- Lightweight K8s ML Proxy and Autoscaler
- K8s ML Proxy and Autoscaler

samos123 (changes requested):
Mostly some opinions around naming.

alpe (commented):
Good work

.github/workflows/build-push.yml
nstogner: Screenshot 2024-01-15 at 8 31 22 PM

nstogner: I am fine reverting, but the reason I made these changes: "Build and Push" as the title of the workflow and "container" as what is being built and pushed. I think it displays nicely on the output.

.github/workflows/load-tests.yml
nstogner: See other response

.github/workflows/build-push.yml
nstogner: See other response

README.md
nstogner: What about leaving the title `# Lingo` and keeping the description in the description section of the repo? I feel like these are too long for an H1.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: ">=1.21.0"
+
+      - name: Run unit tests
+        run: make test-unit
```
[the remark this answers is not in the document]
nstogner: Ahh good to know, thanks!

samos123 (approved):
Names look good based on screenshot

PR #61, opened 2024-01-18
Retries - ReverseProxy.ErrorHandler based approach

### LINKED ISSUE #48: Lingo should retry on proxy failure

nstogner:
Allow for retrying failed requests to backends. Default to 1 retry per lingo-request.

Fixes #48 

Builds on work done by @alpe in #64 and #51.


nstogner:
@alpe here is a variation of your middleware-based retry approach (#64) that does not require wrapping the `http.ResponseWriter`. Would love to get your initial thoughts (still have some work to do).

pkg/proxy/request.go
```diff
@@ -0,0 +1,113 @@
...
+	pr.model = payload.Model
+
+	if pr.model == "" {
+		return fmt.Errorf("no model specified")
+	}
+
+	return nil
+}
+
+func (pr *proxyRequest) sendErrorResponse(w http.ResponseWriter, status int, format string, args ...interface{}) {
```
alpe, line 84: A helper function for writing errors makes a lot of sense. I had this in my head for some time already. ;-)
In the implementation you don't differentiate between log and the error message returned. A log is usually more verbose and may contain internals that you don't want to share with anonymous clients.

pkg/proxy/handler.go
```diff
@@ -158,5 +131,37 @@ func newReverseProxy(host string) *httputil.ReverseProxy {
...
+	proxy.ModifyResponse = func(r *http.Response) error {
+		if h.isRetryCode(r.StatusCode) {
+			// Returning an error will trigger the ErrorHandler.
+			return errors.New("retry")
+		}
+		return nil
+	}
+
+	proxy.ErrorHandler = func(w http.ResponseWriter, r *http.Request, err error) {
+		if err != nil && pr.attempt < h.MaxRetries {
```
alpe, line 144: This retries on any error. Is this correct?

pkg/proxy/handler.go
```diff
@@ -158,5 +131,37 @@ func newReverseProxy(host string) *httputil.ReverseProxy {
...
+			log.Println("retrying")
+			pr.attempt++
+			h.proxyHTTP(w, pr)
+			return
+		}
+
+		pr.sendErrorResponse(w, http.StatusBadGateway, "proxy: exceeded retries: %v/%v", pr.attempt, h.MaxRetries)
```
alpe, line 151: IMHO the error message contains details that should not be shared. The retry mechanism is vulnerable for DoS attacks.

pkg/proxy/handler.go
```diff
@@ -1,153 +1,126 @@
...
-	captureStatusRespWriter := newCaptureStatusCodeResponseWriter(w)
-	w = captureStatusRespWriter
-	timer := prometheus.NewTimer(prometheus.ObserverFunc(func(v float64) {
-		httpDuration.WithLabelValues(modelName, strconv.Itoa(captureStatusRespWriter.statusCode)).Observe(v)
-	}))
-	defer timer.ObserveDuration()
-
-	id := uuid.New().String()
-	log.Printf("request: %v", r.URL)
+	log.Printf("url: %v", r.URL)
```
alpe, line 56: this method is much cleaner and easier to read without the error response handling 👍

alpe (commented):
Very nice start and I am happy to see a fresh approach on the retry problem.

pkg/proxy/request.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func (pr *proxyRequest) setStatus(w http.ResponseWriter, code int) {
+	pr.status = code
+	w.WriteHeader(code)
+}
+
+func (pr *proxyRequest) httpRequest() *http.Request {
+	clone := pr.r.Clone(pr.r.Context())
+	clone.Body = io.NopCloser(bytes.NewReader(pr.body))
```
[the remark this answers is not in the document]
nstogner: Good catch, I'll look at the TeeReader approach. I think we could also do a nil check on `pr.body` before resetting `close.Body` here as well.

pkg/proxy/request.go
nstogner: I was tied between this and splitting the logging out completely. What I landed on was the `if status >= 500` check below that will filter out any messages on internal errors.

pkg/proxy/request.go
```diff
@@ -0,0 +1,113 @@
...
+
+}
+
+func (p *proxyRequest) done() {
+	p.timer.ObserveDuration()
+}
+
+func (pr *proxyRequest) parseModel() error {
+	pr.model = pr.r.Header.Get("X-Model")
+	if pr.model != "" {
```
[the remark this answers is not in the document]
nstogner: Good catch, need to account for this.

pkg/proxy/handler.go
nstogner: This error message will not reach the client because the status code is `> 500`... See the convo here: https://github.com/substratusai/lingo/pull/61/files/a9bc8b12c6b49198b1799bf0ecc0763ab60c680a#r1464478674 ... Perhaps this is a little unexpected and should be refactored. Agreed, retry mechanism can create more load than needed. I think DoS attacks can be mitigated at ingress layers before they reach Lingo. Thoughts?

pkg/proxy/handler.go
nstogner: My understanding is this will retry on any error related to reaching / connecting to the backend by default (for example errors returned by `http.Get()`). For status code errors it will only retry on those reported in `.ModifyResponse`. I think this is the desired behavior. Thoughts?

pkg/proxy/request.go
nstogner: Would like to get your take on this approach.

pkg/proxy/handler.go
```diff
@@ -24,20 +21,34 @@ func MustRegister(r prometheus.Registerer) {
...
 // It is also responsible for triggering scale-from-zero.
 type Handler struct {
-	Deployments *deployments.Manager
-	Endpoints   *endpoints.Manager
-	Queues      *queue.Manager
-	MaxRetries  int
-	RetryCodes  map[int]struct{}
+	Deployments DeploymentManager
+	Endpoints   EndpointManager
+	Queues      QueueManager
```
alpe, line 42: good use of interfaces 👍

pkg/proxy/handler_test.go
```diff
@@ -1,69 +1,157 @@
...
+			expBackendRequestCount: 1,
+		},
+		"retryable 500": {
+			reqMethod:   http.MethodPost,
+			reqPath:     "/",
+			reqBody:     fmt.Sprintf(`{"model":%q}`, model1),
+			backendCode: http.StatusInternalServerError,
+			backendBody: `{"err":"oh no!"}`,
+			expCode:     http.StatusBadGateway,
+			expBody:     `{"error":"Bad Gateway"}` + "\n",
```
alpe, line 89: When retry is enabled, the original backend error msg is never passed to the client. This can be a valid design decision but we should be sensible what error codes are "retryable".  Do you have some scenarios for `500` in mind that can succeed on a retry?

pkg/proxy/handler_test.go
```diff
@@ -1,69 +1,157 @@
...
+		model1: "deploy1",
+		model2: "deploy2",
+	}
+
 	specs := map[string]struct {
-		request   *http.Request
-		expCode   int
-		expLabels map[string]string
+		reqMethod string
+		reqPath   string
```
alpe, line 33: personal preference: reqMethod + reqPath are the same in all tests. It would be more readable to move them to the call and reduce the number of options in the test.

alpe (commented):
Very nice tests! 🏄 
It would be good to have the `X-Model` header scenario covered, when implemented

pkg/proxy/request.go
alpe: I missed the condition, sorry. Let's give it a try

pkg/proxy/handler.go
alpe: the doc specifies it `that handles errors reaching the backend or errors from ModifyResponse.` Looks good

pkg/proxy/request.go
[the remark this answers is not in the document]
nstogner: Took the approach of checking for nil

pkg/proxy/handler_test.go
nstogner: Good catch, we should probably only filter 500s

nstogner: Scratch that, I am going to just pipe through the response in the case where it got an http response by filtering out retry errors.

nstogner: Updated

pkg/proxy/handler_test.go
nstogner: Agreed

pkg/proxy/request.go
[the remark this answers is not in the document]
nstogner: Also added a test case for making sure the right response body was reaching the backend.

pkg/proxy/request.go
```diff
@@ -0,0 +1,126 @@
...
+	pr.status = code
+	w.WriteHeader(code)
+}
+
+// httpRequest returns a new http.Request that is a clone of the original
+// request, preserving the original request body even if it was already
+// read (i.e. if the body was inspected to determine the model).
+func (pr *proxyRequest) httpRequest() *http.Request {
+	clone := pr.r.Clone(pr.r.Context())
+	if pr.body != nil {
```
alpe, line 122: good to check

pkg/proxy/handler.go
```diff
@@ -158,5 +145,50 @@ func newReverseProxy(host string) *httputil.ReverseProxy {
...
+		}
+
+		return nil
+	}
+
+	proxy.ErrorHandler = func(w http.ResponseWriter, r *http.Request, err error) {
+		// This point could be reached if a bad response code was sent by the backend
+		// or
+		// if there was an issue with the connection and no response was ever received.
+		if err != nil && pr.attempt < h.MaxRetries {
```
alpe, line 166: It would be good to abort early when the request context is canceled. I have debugged retries in #65 that never hit a backend
```suggestion
		if err != nil &&
			r.Context().Err() == nil &&
			pr.attempt < h.MaxRetries {
```

pkg/proxy/handler.go
```diff
@@ -158,5 +145,50 @@ func newReverseProxy(host string) *httputil.ReverseProxy {
...
+		if !errors.Is(err, ErrRetry) {
+			pr.sendErrorResponse(w, http.StatusBadGateway, "proxy: exceeded retries: %v/%v", pr.attempt, h.MaxRetries)
+		}
+	}
+
+	log.Printf("Proxying request to host %v: %v\n", host, pr.id)
+	proxy.ServeHTTP(w, pr.httpRequest())
+}
+
+var ErrRetry = errors.New("retry")
```
alpe, line 183: 👍 to have this as an error type

alpe (commented):
Good progress!
As elaborated on discord, the `X-Model` header scenario is not fully handled for retries. The request body should not be empty.

nstogner:
Added a test case in the handler tests for the failure you mentioned. Seeing it now.

pkg/proxy/handler.go
nstogner: It looks like there are checks for the context being done in the reverse proxy .ServeHTTP function. Do you know what cases it would fix to also do the check here?

pkg/proxy/request.go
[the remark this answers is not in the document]
nstogner: I think this is fixed now. I included a test case for this.

pkg/proxy/request.go
```diff
@@ -0,0 +1,127 @@
...
+	pr.timer = prometheus.NewTimer(prometheus.ObserverFunc(func(v float64) {
+		httpDuration.WithLabelValues(pr.model, strconv.Itoa(pr.status)).Observe(v)
+	}))
+
+	return pr
+
+}
+
+// done should be called when the original client request is complete.
+func (p *proxyRequest) done() {
```
samos123, line 54: nit, make it consistent and use pr here too

samos123 (commented):
overall LGTM, could you rebase on main so the additional system tests get triggered?

pkg/proxy/handler_test.go
```diff
@@ -0,0 +1,259 @@
...
+			expCode:      http.StatusBadGateway,
+			expBody:      `{"error":"Bad Gateway"}` + "\n",
+			expLabels: map[string]string{
+				"model":       model1,
+				"status_code": "502",
+			},
+			expBackendRequestCount: 1 + maxRetries,
+		},
+	}
+	for name, spec := range specs {
```
alpe, line 136: good test!

alpe (approved):
Thanks for the updates! The PR look good now and covers all the scenarios discussed . 👍 Great work

pkg/proxy/handler.go
```diff
@@ -145,18 +147,22 @@ func (h *Handler) proxyHTTP(w http.ResponseWriter, pr *proxyRequest) {
...
+
 		return nil
 	}
 
 	proxy.ErrorHandler = func(w http.ResponseWriter, r *http.Request, err error) {
 		if err != nil && pr.attempt < h.MaxRetries {
-			log.Println("retrying")
 			pr.attempt++
+
+			log.Printf("Retrying request (%v/%v): %v", pr.attempt, h.MaxRetries, pr.id)
```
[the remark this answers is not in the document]
samos123: agree, we should refactor all logging statements. I've filed a bug here: https://github.com/substratusai/lingo/issues/79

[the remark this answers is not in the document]
nstogner: Going to defer all of these until refactoring to structured logging

Issue #49, opened 2024-01-08
Improve supply chain security


alpe:
- [ ] use [Harden Runner](https://github.com/step-security/harden-runner) in all GH workflows
- [ ] use hashes instead of versions in GH workflows
- [ ] add [OpenSSF Scorecard](https://github.com/ossf/scorecard) GH workflow to collect issues
- [ ] use cosigner to sign docker image via GH action
- [ ] use [dependency-review](https://github.com/actions/dependency-review-action) GH action
- [ ] add [security.md](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) with contact details

I found some nice examples in https://github.com/sozercan/aikit/tree/main/.github/workflows

Not related to supply chain security but code quality
- [ ] add codeql GH action
- [ ] add linter GH action

samos123:
I'm all for security improvements, but at the same need to ensure that:
* doesn't make it more painful to do releases and development on the project
* doesn't hurt the UX
* doesn't significantly increase complexity

I think all of your tasks make sense, except Harden Runner might make adding and updating GH workflows a bit more painful, but at the same time it does seem good to prevent malicious PRs as well.

PR #62, opened 2024-01-19
Fix issues reported by gosec

alpe:
See #49 
Fixes issues found by a [gosec](https://github.com/securego/gosec) run

* Introduces "request-header-timeout" param


Makefile
```diff
@@ -26,3 +26,32 @@ envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
 	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
+GOLANGCI ?= $(LOCALBIN)/golangci-lint
+
+.PHONY: golangci
+golangci: $(GOLANGCI) ## Download golangci-lint locally if necessary.
+$(GOLANGCI): $(LOCALBIN)
```
alpe, line 33: here and L43-46: I do not really know what this is doing. I copied this from envtest but it seems to work as expected

pkg/deployments/scaler.go
```diff
@@ -86,9 +86,9 @@ func (s *scaler) compareScales(current, desired int32) {
 
 	log.Printf("Comparing scales, current: %v, desired: %v", s.currentScale, s.desiredScale)
 
-	if s.desiredScale > s.currentScale {
+	if s.desiredScale > s.currentScale { //nolint:gocritic
```
alpe, line 89: the linter suggests a `switch` statement instead. I kind of agree but this is personal preference

pkg/deployments/manager.go
```diff
@@ -244,6 +245,9 @@ func getAnnotationInt32(ann map[string]string, key string, defaultValue int32) i
 		log.Printf("parsing annotation as int: %v", err)
 		return defaultValue
 	}
-
-	return int32(value)
+	if value > math.MaxInt32 {
```
alpe, line 248: this issue was highlighted by gosec to prevent an overflow

Makefile
```diff
@@ -26,3 +26,32 @@ envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
...
+	test -s $(LOCALBIN)golangci-lint || GOBIN=$(LOCALBIN) go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.55.2
+
+GCI ?= $(LOCALBIN)/gci
+
+.PHONY: formatter
+formatter: $(GCI) ## Download gci locally if necessary.
```
nstogner, line 51: I would prefer to avoid depending on an external tool in someone's personal repo unless it brings a lot of benefit. I dont think that import ordering is worth the exposure.

cmd/lingo/main.go
```diff
@@ -154,19 +154,23 @@ func run() error {
...
+	statsServer := &http.Server{Addr: ":8083", Handler: statsHandler, ReadHeaderTimeout: requestHeaderTimeout}
 
 	var wg sync.WaitGroup
 	wg.Add(1)
 	go func() {
 		defer func() {
-			statsServer.Shutdown(context.Background())
-			proxyServer.Shutdown(context.Background())
+			if err := statsServer.Shutdown(context.Background()); err != nil {
+				setupLog.Error(err, "shutdown stats server")
```
nstogner, line 169: This will log an error under normal shutdown circumstances. https://pkg.go.dev/net/http#Server.Shutdown

pkg/autoscaler/autoscaler.go
```diff
@@ -70,21 +71,21 @@ func (r *Autoscaler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	return ctrl.Result{}, nil
 }
 
-func (a *Autoscaler) Start() {
-	for range time.Tick(a.Interval) {
-		if !a.LeaderElection.IsLeader.Load() {
+func (r *Autoscaler) Start() {
```
nstogner, line 74: Why rename to `r`?

pkg/proxy/handler.go
```diff
@@ -56,7 +56,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		modelName = "unknown"
 		log.Printf("error reading model from request body: %v", err)
 		w.WriteHeader(http.StatusBadRequest)
-		w.Write([]byte("Bad request: unable to parse .model from JSON payload"))
+		_, _ = w.Write([]byte("Bad request: unable to parse .model from JSON payload"))
```
nstogner, line 59: I this required to make the linter happy?

pkg/queue/queue.go
```diff
@@ -152,7 +153,10 @@ func (q *Queue) Start() {
 			continue
 		}
 
-		itm := e.Value.(*item)
```
nstogner, line 155: Panic was implicit here right? Why the explicit call to panic? Linter?

nstogner (commented):
A useful PR. Thanks for submitting! Added some comments.

Makefile
alpe: Import order can be annoying in PR reviews and may cause merge conflicts. I prefer standardization but this change was mainly driven by the linter. We would need to remove the [`cgi`](https://github.com/substratusai/lingo/pull/62/files#diff-9917ddc9f1c3304218f7269265b746d997c5c0615478177b5fceecd33ef47cb5R29) rule

pkg/autoscaler/autoscaler.go
alpe: The name is `r` for the other methods. The linter highlighted the inconsistency.

pkg/proxy/handler.go
alpe: yes, either this way or with a `//nolint` annotation. The rule is that every error is handled or explicit ignored.

pkg/queue/queue.go
alpe: This change is for the linter only. Similar to the error rule, somebody should make the decision to handle this. An alternative solution is the `//nolint` comment.

Makefile
samos123: I think CGI is fine, it has a lot of contributors and is used in a lot of projects: https://github.com/daixiang0/gci

cmd/lingo/main.go
```diff
@@ -71,11 +71,13 @@ func run() error {
 	var metricsAddr string
 	var probeAddr string
 	var concurrencyPerReplica int
+	var requestHeaderTimeout time.Duration
 
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8082", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.IntVar(&concurrencyPerReplica, "concurrency", concurrency, "the number of simultaneous requests that can be processed by each replica")
 	flag.IntVar(&scaleDownDelay, "scale-down-delay", scaleDownDelay, "seconds to wait before scaling down")
+	flag.DurationVar(&requestHeaderTimeout, "request-header-timeout", 10*time.Second, "amount of time for the client to send headers before a timeout error will occur")
```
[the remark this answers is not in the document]
samos123: I think this would break scale from 0 with gpus? sometimes scale up takes 10+ minutes and a request will just be hanging for 10 minutes until the backend comes up.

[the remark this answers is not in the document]
samos123: I would vote for setting it to 30 minutes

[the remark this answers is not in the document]
samos123: I am actually not sure how ReadHeaderTimeout behaves. Does the proxy/handler first return headers before returning the full body and will it return headers before the backend is up?

If the answer is yes, then I think current value of 10 seconds is fine. I was reading this to improve my understanding: https://adam-p.ca/blog/2022/01/golang-http-server-timeouts/

pkg/autoscaler/autoscaler.go
samos123: `a` seems to make more sense for `Autoscaler`. I think the others might have used `r` because they were reconcilsers?

samos123 (commented):
LGTM, but unsure about the RequestHeaderTimeout.

cmd/lingo/main.go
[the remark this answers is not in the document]
alpe: The read header timeout is configured for the lingo http + stats server only and does not affect the proxy or backend timeouts.
This setting protects against slow clients that do not complete their requests. See [slowloris attack](https://medium.com/a-journey-with-go/go-understand-and-mitigate-slowloris-attack-711c1b1403f6) for more context.

samos123:

PR #64, opened 2024-01-22
Use retry middleware with reverse proxy

alpe:
Same middleware as in #51  but set up for the reverse proxy

### DIFF
--- a/pkg/proxy/handler.go
+++ b/pkg/proxy/handler.go
@@ -123,22 +123,6 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	newReverseProxy(host).ServeHTTP(w, proxyRequest)
 }
 
-func withInflightCounted(endpoints *endpoints.Manager, deploy string, host string) func(other http.Handler) http.HandlerFunc {
-	return func(other http.Handler) http.HandlerFunc {
-		return func(w http.ResponseWriter, r *http.Request) {
-			done, err := endpoints.RegisterInFlight(deploy, host)
-			if err != nil {
-				log.Printf("error registering in-flight request: %v", err)
-				w.WriteHeader(http.StatusInternalServerError)
-				return
-			}
-			defer done()
-
-			other.ServeHTTP(w, r)
-		}
-	}
-}
-
 // parseModel parses the model name from the request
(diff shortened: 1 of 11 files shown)


pkg/proxy/handler.go
```diff
@@ -105,10 +115,19 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
+	defer done()
+
+	log.Printf("Proxying request to host %v: %v\n", host, id)
 	// TODO: Avoid creating new reverse proxies for each request.
 	// TODO: Consider implementing a round robin scheme.
-	log.Printf("Proxying request to host %v: %v\n", host, id)
-	newReverseProxy(host).ServeHTTP(w, proxyRequest)
+	proxy := newReverseProxy(host)
+	NewRetryMiddleware(h.retriesOnErr, proxy).ServeHTTP(w, proxyRequest)
```
alpe, line 130: This is how the middleware is integrated with the reverse proxy

pkg/proxy/middleware.go
```diff
@@ -0,0 +1,282 @@
...
+	headerBuf      http.Header
+	wroteHeader    bool
+	statusCode     int
+	immediateWrite bool
+	isDiscardable  func(status int) bool
+}
+
+// NewDiscardableResponseWriter creates a new instance of the response writer delegator.
+// It takes a http.ResponseWriter and a function to determine by the status code if content should be written or discarded (for retry).
+func NewDiscardableResponseWriter(writer http.ResponseWriter, isDiscardable func(status int) bool) XResponseWriter {
```
nstogner, line 127: I think wrapping the http.ResponseWriter comes with thorns. You can lose functionality if you don't implement all of the interfaces that ResponseWriter can be: https://go.dev/play/p/3B0yYC8kTvf

This is one the reasons I think that the middleware pattern in Go typically leads to more code than necessary. In the case of Lingo where we have a single primary handler (the proxy) I think that the middleware approach leads to more a more verbose/complicated solution than leaning on helper functions.

nstogner: Bringing benefits being customizable discussion here: https://github.com/substratusai/lingo/issues/63

PR #65, opened 2024-01-31
Spike: Add integration test

alpe:
Some test scenarios


nstogner (commented):
I hesitate to add these tests as I think we have coverage for all of these code paths covered with the proxy tests and the existing integration tests.

alpe:
No worries. I created these tests for the middleware and used them to confirm/ showcase test scenarios. Your tests look good

Issue #67, opened 2024-02-05
Flapping scale from 0 to 1 to 0 to 1


samos123:
I am not sure how exactly  to trigger it, but I can frequently trigger this with replicas = 3, not 100% of the time though.

Sharing some info:
```
k get pods
NAME                                  READY   STATUS        RESTARTS   AGE
lingo-784b5857fc-9vz27                1/1     Running       0          9m18s
lingo-784b5857fc-hssfj                1/1     Running       0          9m18s
lingo-784b5857fc-pnx87                1/1     Running       0          9m18s
stapi-minilm-l6-v2-5dd7bb4fc8-mw9nr   1/1     Terminating   0          41s
stapi-minilm-l6-v2-5dd7bb4fc8-svb55   0/1     Running       0          11s
```

Events on the deployment that demonstrate 12 x over last 5 minute the deployment was scaled up and down. However, no requests have been in the queue or been sent to lingo during this time.
```
  Normal  ScalingReplicaSet  25s (x12 over 5m58s)  deployment-controller  Scaled down replica set stapi-minilm-l6-v2-5dd7bb4fc8 to 0 from 1
  Normal  ScalingReplicaSet  25s (x12 over 5m58s)  deployment-controller  Scaled up replica set stapi-minilm-l6-v2-5dd7bb4fc8 to 1 from 0
```

It seems there are 2 lingo replicas at the same time that make conflicting decisions.

Replica 1, lingo-784b5857fc-hssfj:
```
2024/02/05 05:48:59 Comparing scales, current: 1, desired: 1
2024/02/05 05:49:02 Not leader, doing nothing
2024/02/05 05:49:05 Not leader, doing nothing
2024/02/05 05:49:08 Not leader, doing nothing
2024/02/05 05:49:11 Not leader, doing nothing
2024/02/05 05:49:14 Not leader, doing nothing
2024/02/05 05:49:14 UpdateState(0, 0, 3)
2024/02/05 05:49:14 Comparing scales, current: 0, desired: 1
2024/02/05 05:49:14 Scaling model "stapi-minilm-l6-v2": 1
2024/02/05 05:49:14 UpdateState(1, 0, 3)
2024/02/05 05:49:14 Comparing scales, current: 1, desired: 1
2024/02/05 05:49:14 Updating queue size: deployment: stapi-minilm-l6-v2, replicas: 0, newSize: 100
2024/02/05 05:49:14 UpdateState(1, 0, 3)
[...]

PR #70, opened 2024-02-06
Stop scale down timer

alpe:
See: #67 

When a deployment is deleted on reconcile there may still be a scale down timer running that modifies k8s replica state later. With this PR, the timer is stopped.

### DIFF
--- a/pkg/deployments/manager_test.go
+++ b/pkg/deployments/manager_test.go
@@ -144,58 +144,49 @@ func TestAddDeployment(t *testing.T) {
 func TestRemoveDeployment(t *testing.T) {
 	const myDeployment = "myDeployment"
 	specs := map[string]struct {
-		setup  func(t *testing.T, m *Manager)
-		assert func(t *testing.T, m *Manager)
+		setup      func(t *testing.T, m *Manager)
+		delay      time.Duration
+		expScalers map[string]scale
 	}{
 		"single model deployment": {
 			setup: func(t *testing.T, m *Manager) {
-				m.modelToDeployment["model1"] = myDeployment
-				m.scalers[myDeployment] = &scaler{}
+				m.setModelMapping("model1", myDeployment)
+				m.getScaler(myDeployment)
 			},
-			assert: func(t *testing.T, m *Manager) {
-				assert.Len(t, m.modelToDeployment, 0)
-				assert.Len(t, m.scalers, 0)
-			},
+			expScalers: map[string]scale{},
 		},
 		"multi model deployment": {
 			setup: func(t *testing.T, m *Manager) {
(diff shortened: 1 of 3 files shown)


pkg/deployments/scaler.go
```diff
@@ -116,6 +116,15 @@ func (s *scaler) compareScales(current, desired int32) {
 	}
 }
 
+func (s *scaler) StopScaleDownTimer() {
```
nstogner, line 119: I would prefer to have a general `.Stop()` method on scaler if scaler needs some cleanup to be done before deletion.

pkg/deployments/manager_test.go
```diff
@@ -136,6 +141,79 @@ func TestAddDeployment(t *testing.T) {
 	}
 }
 
+func TestRemoveDeployment(t *testing.T) {
```
nstogner, line 144: I would prefer that we write these tests utilizing as many of the exported methods as possible and written in a manner that follows expected usage. This allows the tests to be more flexible over time. One example: asserting on `m.ResolveDeployment()` as opposed to `m.modelToDeployment`. I also think this is a good case for not using table-driven tests. In this case a table driven design is really only reusing `m.removeDeployment(req)` across all of the tests but it becomes harder to read and covers less surface area per lines of test than a usage/story-driven approach.

nstogner (changes requested):

pkg/deployments/manager_test.go
alpe: Thanks for the feedback! I can see the benefit of using method rather than messing with internal state. I also agree that the assert function is not great and values should be stored in the test spec. I like table test as a spec for the method. But I am so used to writing them, that I have a hard time to think how this test would look like in a classic fashion. Would you split it into multiple separate tests or work with a single manager and add/ remove state?

samos123:
This won't fix #67 since in my case I never deleted a deployment yet still triggered the issue. So not sure if this patch is needed. That being said, I think this fix is helpful because otherwise the timers would just keep on going even after the scaler is deleted?

This was an interesting read: https://medium.com/@oboturov/golang-time-after-is-not-garbage-collected-4cbc94740082

samos123 (approved):

pkg/deployments/manager_test.go
samos123: I personally like the table tests as well but I like to limit them to data only and not logic. I think as a follow up we need to have tests for scaler.go and refactor it before we can fix #67 which is imo the highest priority issue right now.

If we think these tests on internal state are no longer productive, then we can simply remove the tests later as well. The issue with testing using internal state is that it makes refactoring the implementation harder later.

samos123:
I will include this in #76

PR #74, opened 2024-02-07
[Do not merge]scaler reset desired scale

alpe:
Demo, do not merge
See #67

### DIFF
--- a/pkg/deployments/scaler.go
+++ b/pkg/deployments/scaler.go
@@ -85,35 +85,38 @@ func (s *scaler) compareScales(current, desired int32) {
 	if s.desiredScale > s.currentScale {
 		// Scale up immediately.
 		go s.scaleFunc(s.desiredScale, false)
-		s.scaleDownStarted = false
+		s.stopScaleDown()
 	} else if s.desiredScale == s.currentScale {
 		// Do nothing, schedule nothing.
-		if s.scaleDownTimer != nil {
-			s.scaleDownTimer.Stop()
-		}
-		s.scaleDownStarted = false
+		s.stopScaleDown()
 	} else {
 		// Schedule a scale down.
-
 		if s.scaleDownTimer == nil {
 			s.scaleDownTimer = time.AfterFunc(s.scaleDownDelay, func() {
-				if s.desiredScale == -1 || s.desiredScale == s.currentScale {
-					s.scaleDownStarted = false
+				s.mtx.Lock()
+				s.scaleDownStarted = false // mark completed already
+				desiredScale, currentScale := s.desiredScale, s.currentScale
+				s.mtx.Unlock()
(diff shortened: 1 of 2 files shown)


pkg/deployments/scaler.go
```diff
@@ -33,12 +33,15 @@ func (s *scaler) AtLeastOne() {
...
 func (s *scaler) UpdateState(replicas, min, max int32) {
 	log.Printf("UpdateState(%v, %v, %v)", replicas, min, max)
-	s.setMinMax(min, max)
-	s.compareScales(replicas, -1)
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+	s.minScale = min
+	s.maxScale = max
+	s.currentScale = replicas
+	s.desiredScale = -1
```
alpe, line 44: This line clears the desired scale and is called by reconcile. This resets the state on all instances and fixes the stale state problem for ex-leader instances. Nevertheless this is not ideal for the current leader as the desired state may be set already.

pkg/deployments/scaler.go
```diff
@@ -89,33 +85,41 @@ func (s *scaler) compareScales(current, desired int32) {
 	if s.desiredScale > s.currentScale {
 		// Scale up immediately.
 		go s.scaleFunc(s.desiredScale, false)
-		s.scaleDownStarted = false
+		s.stopScaleDown()
```
alpe, line 88: ensures that scale down is stopped

pkg/deployments/scaler.go
```diff
@@ -89,33 +85,41 @@ func (s *scaler) compareScales(current, desired int32) {
...
 	} else {
 		// Schedule a scale down.
-
 		if s.scaleDownTimer == nil {
 			s.scaleDownTimer = time.AfterFunc(s.scaleDownDelay, func() {
-				if err := s.scaleFunc(s.desiredScale, false); err != nil {
+				s.mtx.Lock()
+				s.scaleDownStarted = false // mark completed already
```
alpe, line 97: also set to false for the error case

pkg/deployments/scaler.go
```diff
@@ -89,33 +85,40 @@ func (s *scaler) compareScales(current, desired int32) {
...
+				desiredScale, currentScale := s.desiredScale, s.currentScale
+				s.mtx.Unlock()
+				if desiredScale == -1 || desiredScale == currentScale {
+					return
+				}
+				if err := s.scaleFunc(desiredScale, false); err != nil {
 					log.Printf("task: run error: %v", err)
-				} else {
-					s.scaleDownStarted = false
-					s.compareScales(s.desiredScale, -1)
```
alpe, line 108: Not sure why does this call exists? The desired state should be set already.

PR #75, opened 2024-02-07
[Do not merge] alternative approach to clear state when not leader anymore

alpe:
Based on #70 

The autoscaler registers a callback on the elector when to stop scalers when not leader anymore.
This PR also contains changes to the scaler from #74

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -71,6 +71,7 @@ func (r *Autoscaler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 }
 
 func (a *Autoscaler) Start() {
+	a.LeaderElection.AfterOnStoppedLeading(a.Deployments.StopScalers)
 	for range time.Tick(a.Interval) {
 		if !a.LeaderElection.IsLeader.Load() {
 			log.Println("Not leader, doing nothing")
--- a/pkg/deployments/scaler.go
+++ b/pkg/deployments/scaler.go
@@ -116,7 +116,8 @@ func (s *scaler) compareScales(current, desired int32) {
 	}
 }
 
-func (s *scaler) StopScaleDownTimer() {
+// Stop stops the scale down process for the scaler.
+func (s *scaler) Stop() {
 	s.mtx.Lock()
 	defer s.mtx.Unlock()
 	if s.scaleDownTimer != nil {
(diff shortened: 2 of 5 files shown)


pkg/autoscaler/autoscaler.go
```diff
@@ -71,6 +71,7 @@ func (r *Autoscaler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 }
 
 func (a *Autoscaler) Start() {
+	a.LeaderElection.AfterOnStoppedLeading(a.Deployments.StopScalers)
```
alpe, line 74: cleanup desired scale and downscale when instance is not leader anymore

pkg/deployments/scaler.go
```diff
@@ -89,19 +89,22 @@ func (s *scaler) compareScales(current, desired int32) {
...
 		// Schedule a scale down.
-
 		if s.scaleDownTimer == nil {
 			s.scaleDownTimer = time.AfterFunc(s.scaleDownDelay, func() {
-				if err := s.scaleFunc(s.desiredScale, false); err != nil {
+				s.mtx.Lock()
+				s.scaleDownStarted = false // mark completed already
+				desiredScale, currentScale := s.desiredScale, s.currentScale
+				s.mtx.Unlock()
+				if desiredScale == -1 || desiredScale == currentScale {
```
samos123, line 104: s.desiredScale could only be `-1` right after initialization right? because there is this if earlier in scaler:

```
	if current != -1 {
		s.currentScale = current
	}
	if desired != -1 {
		s.desiredScale = desired
	}
```

asking to improve my understanding, this is not feedback to your code or a request for any changes

pkg/autoscaler/autoscaler.go
samos123: I think this would break scale from 0 in the proxy which should still be possible

samos123 (commented):
changes make sense overall except stopping all scalers, I think that would break the `h.Deployments.AtLeastOne(deploy)` call in the proxies that are not a leader.

pkg/autoscaler/autoscaler.go
samos123: I think this would break the `h.Deployments.AtLeastOne(deploy)` call in the proxies that are not a leader.

samos123: nvm it won't break that since it's a different flow. Currently all stop does is to stop the scale down timer.

samos123:
will include this in #76

PR #76, opened 2024-02-07
Fix scale back to 0 with 300 requests scenario

### LINKED ISSUE #73: Scale to 0 not working with replicas 3
```
2024/02/07 07:11:45 SetDesiredScale(0)                                                                          
2024/02/07 07:11:45 Is leader, autoscaling                                                                      
2024/02/07 07:11:45 Average for deployment: lingo: 0 (ceil: 0), current wait count: 0       
2024/02/07 07:11:45 SetDesiredScale(0)                                                                          
2024/02/07 07:11:45 Is leader, autoscaling                                                                      
2024/02/07 07:11:45 Average for deployment: stapi-minilm-l6-v2: 0 (ceil: 0), current wait count: 0              
2024/02/07 07:11:45 SetDesiredScale(0)                                                                          
2024/02/07 07:11:45 Comparing scales, current: 2, desired: 0                                                    
2024/02/07 07:11:48 Calculating scales for all                                                                  
2024/02/07 07:11:48 Leader active requests for: lingo: 0                                          
2024/02/07 07:11:48 Leader active requests for: stapi-minilm-l6-v2: 0                                           
2024/02/07 07:11:48 Leader active requests for: kubernetes: 0                               
2024/02/07 07:11:48 Aggregating active requests for endpoint: 10.244.0.6:8083: kubernetes: 0
2024/02/07 07:11:48 Aggregating active requests for endpoint: 10.244.0.6:8083: lingo: 0   
2024/02/07 07:11:48 Aggregating active requests for endpoint: 10.244.0.6:8083: stapi-minilm-l6-v2: 0
2024/02/07 07:11:48 Aggregating active requests for endpoint: 10.244.0.7:8083: kubernetes: 0
2024/02/07 07:11:48 Aggregating active requests for endpoint: 10.244.0.7:8083: lingo: 0
2024/02/07 07:11:48 Aggregating active requests for endpoint: 10.244.0.7:8083: stapi-minilm-l6-v2: 0
2024/02/07 07:11:48 Is leader, autoscaling                                                                      
2024/02/07 07:11:48 Average for deployment: lingo: 0 (ceil: 0), current wait count: 0               
2024/02/07 07:11:48 SetDesiredScale(0)                                                                          
2024/02/07 07:11:48 Is leader, autoscaling                                                                      
2024/02/07 07:11:48 Average for deployment: stapi-minilm-l
2024/02/07 07:11:48 SetDesiredScale(0)
2024/02/07 07:11:51 Calculating scales for all
2024/02/07 07:11:51 Leader active requests for: kubernetes: 0
2024/02/07 07:11:51 Leader active requests for: lingo: 0
2024/02/07 07:11:51 Leader active requests for: stapi-minilm-l6-v2: 0
2024/02/07 07:11:51 Aggregating active requests for endpoint: 10.244.0.7:8083: stapi-minilm-l6-v2: 0
2024/02/07 07:11:51 Aggregating active requests for endpoint: 10.244.0.7:8083: kubernetes: 0
2024/02/07 07:11:51 Aggregating active requests for endpoint: 10.244.0.7:8083: lingo: 0
2024/02/07 07:11:51 Aggregating active requests for endpoint: 10.244.0.6:8083: lingo: 0
2024/02/07 07:11:51 Aggregating active requests for endpoint: 10.244.0.6:8083: stapi-minilm-l6-v2: 0
2024/02/07 07:11:51 Aggregating active requests for endpoint: 10.244.0.6:8083: kubernetes: 0
2024/02/07 07:11:51 Is leader, autoscaling
2024/02/07 07:11:51 Average for deployment: kubernetes: 0 (ceil: 0), current wait count: 0
2024/02/07 07:11:51 SetDesiredScale(0)
2024/02/07 07:11:51 Is leader, autoscaling
2024/02/07 07:11:51 Average for deployment: lingo: 0 (ceil: 0), current wait count: 0
2024/02/07 07:11:51 SetDesiredScale(0)
2024/02/07 07:11:51 Is leader, autoscaling
2024/02/07 07:11:51 Average for deployment: stapi-minilm-l6-v2: 0 (ceil: 0), current wait count: 0
2024/02/07 07:11:51 SetDesiredScale(0)
2024/02/07 07:11:51 Comparing scales, current: 2, desired: 0
```

It looks like it should be scaling back to 0 but it just never does, it stays in this state indefinitely.

This is reproducible on main if you run the following:
```
export REPLICAS=3
export REQUESTS=300
export EXPECTED_REPLICAS=2
make test-e2e
```

samos123:
Fixes #73 and includes pr #70 and others from Alex

* Remove compareScales function
* Refactor SetDesiredScale that is simpler to understand by storing lastSuccessfulScale time
* Improve test coverage for scaler.go

Co-author: @alpe

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -130,7 +130,6 @@ func aggregateStats(s stats.Stats, httpc *http.Client, endpoints []string) (stat
 			continue
 		}
 		for k, v := range fetched.ActiveRequests {
-
 			log.Printf("Aggregating active requests for endpoint: %v: %v: %v", endpoint, k, v)
 			s.ActiveRequests[k] = fetched.ActiveRequests[k] + v
 		}
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -71,6 +71,7 @@ func (r *Autoscaler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 }
 
 func (a *Autoscaler) Start() {
+	a.LeaderElection.AfterOnStoppedLeading(a.Deployments.StopScalers)
 	for range time.Tick(a.Interval) {
 		if !a.LeaderElection.IsLeader.Load() {
 			log.Println("Not leader, doing nothing")
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -93,13 +93,6 @@ func (a *Autoscaler) Start() {
 		}
 
 		for deploymentName, waitCount := range stats.ActiveRequests {
-			// TODO remove this check and ensure only stats for deployments with models are returned.
-			// Currently stats.ActiveRequests includes service names / deployments that have no model annotation.
-			// This check is needed to prevent Lingo from scaling itself and other deployments to 0
-			if !a.Deployments.HasModel(deploymentName) {
-				log.Printf("Deployment: %v has no model annotations, skipping", deploymentName)
-				continue
-			}
 			log.Println("Is leader, autoscaling")
 			avg := a.getMovingAvgQueueSize(deploymentName)
 			avg.Next(float64(waitCount))
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -93,6 +93,11 @@ func (a *Autoscaler) Start() {
 		}
 
 		for deploymentName, waitCount := range stats.ActiveRequests {
+			// TODO remove this check and ensure only stats for deployments with models are returned
+			if !a.Deployments.HasModel(deploymentName) {
+				log.Printf("Deployment: %v has no model annotations, skipping", deploymentName)
+				continue
+			}
 			log.Println("Is leader, autoscaling")
 			avg := a.getMovingAvgQueueSize(deploymentName)
 			avg.Next(float64(waitCount))
@@ -130,6 +135,7 @@ func aggregateStats(s stats.Stats, httpc *http.Client, endpoints []string) (stat
 			continue
 		}
 		for k, v := range fetched.ActiveRequests {
+
 			log.Printf("Aggregating active requests for endpoint: %v: %v: %v", endpoint, k, v)
 			s.ActiveRequests[k] = fetched.ActiveRequests[k] + v
 		}
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -93,7 +93,9 @@ func (a *Autoscaler) Start() {
 		}
 
 		for deploymentName, waitCount := range stats.ActiveRequests {
-			// TODO remove this check and ensure only stats for deployments with models are returned
+			// TODO remove this check and ensure only stats for deployments with models are returned.
+			// Currently stats.ActiveRequests includes service names / deployments that have no model annotation.
+			// This check is needed to prevent Lingo from scaling itself and other deployments to 0
 			if !a.Deployments.HasModel(deploymentName) {
 				log.Printf("Deployment: %v has no model annotations, skipping", deploymentName)
 				continue
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -71,7 +71,6 @@ func (r *Autoscaler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 }
 
 func (a *Autoscaler) Start() {
-	a.LeaderElection.AfterOnStoppedLeading(a.Deployments.StopScalers)
 	for range time.Tick(a.Interval) {
 		if !a.LeaderElection.IsLeader.Load() {
 			log.Println("Not leader, doing nothing")
--- a/pkg/deployments/manager.go
+++ b/pkg/deployments/manager.go
@@ -230,6 +230,15 @@ func (r *Manager) ReadinessChecker(_ *http.Request) error {
 	return nil
 }
 
+// StopScalers stops all scheduled scale down processes and resets desired state
+func (r *Manager) StopScalers() {
+	r.scalersMtx.Lock()
+	defer r.scalersMtx.Unlock()
+	for _, s := range r.scalers {
+		s.Stop()
+	}
+}
+
 func getAnnotationInt32(ann map[string]string, key string, defaultValue int32) int32 {
 	if ann == nil {
 		return defaultValue
--- a/pkg/deployments/manager.go
+++ b/pkg/deployments/manager.go
@@ -176,6 +176,10 @@ func (r *Manager) getScalesSnapshot() map[string]scale {
 
 func (r *Manager) scaleFunc(ctx context.Context, deploymentName string) func(int32, bool) error {
 	return func(n int32, atLeastOne bool) error {
+		if !r.HasModel(deploymentName) {
+			return fmt.Errorf("not scaling deployment %q: not managed by this manager", deploymentName)
+		}
+
 		if atLeastOne {
 			log.Printf("Scaling model %q: at-least-one", deploymentName)
 		} else {
--- a/pkg/deployments/manager.go
+++ b/pkg/deployments/manager.go
@@ -40,6 +40,7 @@ type Manager struct {
 
 	ScaleDownPeriod time.Duration
 
+	// TODO remove deploymentsMtx and deployments once #59 is fixed
 	deploymentsMtx sync.RWMutex
 	// deployments is a set of deployment names that are managed by the Manager
 	deployments map[string]struct{}
@@ -113,6 +114,7 @@ func (r *Manager) addDeployment(ctx context.Context, d appsv1.Deployment) error 
 	return nil
 }
 
+// TODO remove this function once #59 is fixed
 func (r *Manager) HasModel(deploymentName string) bool {
 	r.deploymentsMtx.RLock()
 	defer r.deploymentsMtx.RUnlock()
--- a/pkg/deployments/manager.go
+++ b/pkg/deployments/manager.go
@@ -143,6 +143,10 @@ func (r *Manager) removeDeployment(req ctrl.Request) {
 		}
 	}
 	r.modelToDeploymentMtx.Unlock()
+
+	r.deploymentsMtx.Lock()
+	delete(r.deployments, req.Name)
+	r.deploymentsMtx.Unlock()
 }
 
 func (r *Manager) getScaler(deploymentName string) *scaler {
--- a/pkg/deployments/manager.go
+++ b/pkg/deployments/manager.go
@@ -26,7 +26,6 @@ func NewManager(mgr ctrl.Manager) (*Manager, error) {
 	r.Client = mgr.GetClient()
 	r.scalers = map[string]*scaler{}
 	r.modelToDeployment = map[string]string{}
-	r.deployments = map[string]struct{}{}
 	if err := r.SetupWithManager(mgr); err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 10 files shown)


pkg/deployments/scaler.go
```diff
@@ -84,19 +84,22 @@ func (s *scaler) compareScales(current, desired int32) {
...
-		}
-		s.scaleDownStarted = false
+		s.stopScaleDown()
 	} else {
 		// Schedule a scale down.
-
 		if s.scaleDownTimer == nil {
 			s.scaleDownTimer = time.AfterFunc(s.scaleDownDelay, func() {
-				if err := s.scaleFunc(s.desiredScale, false); err != nil {
+				s.mtx.Lock()
```
samos123, line 95: The beginning of the function already has a .Lock and defer .Unlock(). Why do we need to do it again?

alpe: The code is called later in a separate Go routine. A new lock is required therefore

pkg/deployments/scaler.go
```diff
@@ -46,69 +70,53 @@ func (s *scaler) UpdateState(replicas, min, max int32) {
...
-			s.scaleDownTimer.Stop()
-		}
-		s.scaleDownStarted = false
+		log.Println("Desired scale equals current scale, doing nothing")
 	} else {
-		// Schedule a scale down.
-
-		if s.scaleDownTimer == nil {
-			s.scaleDownTimer = time.AfterFunc(s.scaleDownDelay, func() {
+		if time.Since(s.lastScaleDown) >= s.scaleDownDelay {
```
alpe, line 92: 🤔 the `lastScaleDown` value is set in the constructor. This ensures the scale down delay is taken into account on a node startup and right after the last successful scale down. 
There is no delay if a running node becomes leader or if the last scale down is some time ago.  A scale down call would execute immediately as the condition is always true.
With the old logic, the first call to scale down triggered a timer. I order to achieve the same, we can have an unset `lastScaleDown` time and set it on first call to scale down. You can check with `s.lastScaleDown.IsZero()` for unset.
It should be cleared (unset) on 
* desiredScale >= currentScale
* successful scale down
* loosing leader role
* Probably with `AtLeastOne()`

pkg/deployments/manager.go
```diff
@@ -103,6 +113,13 @@ func (r *Manager) addDeployment(ctx context.Context, d appsv1.Deployment) error
 	return nil
 }
 
+func (r *Manager) HasModel(deploymentName string) bool {
+	r.deploymentsMtx.RLock()
+	defer r.deploymentsMtx.RUnlock()
+	_, exists := r.deployments[deploymentName]
+	return exists
+}
+
```
alpe, line 122: Please remove the deployment in `removeDeployment()`

alpe (commented):
Good refactoring and simpler to read. I added some comments as I think the logic has changed for scale down. This is not necessarily bad but should be revisited from a product view.

pkg/autoscaler/autoscaler.go
```diff
@@ -93,6 +93,11 @@ func (a *Autoscaler) Start() {
 		}
 
 		for deploymentName, waitCount := range stats.ActiveRequests {
+			// TODO remove this check and ensure only stats for deployments with models are returned
+			if !a.Deployments.HasModel(deploymentName) {
```
nstogner, line 97: What scenario leads lingo to fall into this state?

pkg/deployments/manager.go
```diff
@@ -39,6 +40,10 @@ type Manager struct {
 
 	ScaleDownPeriod time.Duration
 
+	deploymentsMtx sync.RWMutex
+	// deployments is a set of deployment names that are managed by the Manager
+	deployments map[string]struct{}
```
nstogner, line 45: This appears to be an grow-only map as implemented. Why do we need to track this here?

pkg/deployments/scaler.go
```diff
@@ -14,13 +14,37 @@ type scaler struct {
...
+	scaleDownDelay time.Duration
+}
+
+func newScaler(scaleDownDelay time.Duration, scaleFunc func(int32, bool) error) *scaler {
+	s := &scaler{
+		// -1 represents unknown
+		currentScale:   -1,
+		desiredScale:   -1,
+		scaleDownDelay: scaleDownDelay,
+		lastScaleDown:  time.Now(),
```
nstogner, line 32: Use `time.Time{}` (empty time) for this.

pkg/deployments/scaler.go
```diff
@@ -14,13 +14,37 @@ type scaler struct {
...
 	scaleFuncMtx sync.Mutex
 	scaleFunc    func(n int32, atLeastOne bool) error
 
-	scaleDownDelay   time.Duration
-	scaleDownStarted bool
-	scaleDownTimer   *time.Timer
+	scaleDownDelay time.Duration
+}
+
+func newScaler(scaleDownDelay time.Duration, scaleFunc func(int32, bool) error) *scaler {
```
nstogner, line 26: In future, it would be nice to keep the functions in place to avoid the noisy diff (its hard to know exactly what was changed here)... We can do re-ordering in another PR that does not change logic.

pkg/deployments/scaler.go
```diff
@@ -46,69 +70,53 @@ func (s *scaler) UpdateState(replicas, min, max int32) {
...
-		n = min
-	} else if n > max {
-		n = max
-	}
-	return n
-}
-
-func (s *scaler) compareScales(current, desired int32) {
+	log.Printf("SetDesiredScale(%v), current: %v, min: %v, max: %v", n, s.currentScale, s.minScale, s.maxScale)
+	nMinMax := s.applyMinMax(n)
```
nstogner, line 74: No need for local var, clearer to do `s.desiredScale = s.applyMinMax(n)`

pkg/deployments/scaler.go
nstogner: Here is my code I was working on in parallel:

```go
	if s.currentScale < s.desiredScale {
		// Scale up immediately.
		go s.scaleFunc(s.desiredScale, false)
		s.desiredScaleDownStart = time.Time{}
	} else if s.currentScale == s.desiredScale {
		s.desiredScaleDownStart = time.Time{}
	} else {
		if s.desiredScaleDownStart.IsZero() {
			s.desiredScaleDownStart = time.Now()
		} else if time.Since(s.desiredScaleDownStart) >= s.scaleDownDelay {
			go s.scaleFunc(s.desiredScale, false)
		}
	}
```

pkg/deployments/scaler_test.go
```diff
@@ -74,8 +95,10 @@ func TestSetDesiredScale(t *testing.T) {
 			// Action
 			s.SetDesiredScale(tc.desiredScale)
 
-			// Wait for the scale function to be called
-			mockScaleWG.Wait()
+			if tc.expectedScaleFunc {
+				// Wait for the scale function to be called
+				mockScaleWG.Wait()
```
nstogner, line 100: Prefer to do a conditional `wg.Add(1)` instead of the conditional `.Wait()`.

nstogner (changes requested):

pkg/deployments/scaler.go
```diff
@@ -46,69 +70,53 @@ func (s *scaler) UpdateState(replicas, min, max int32) {
...
-	log.Printf("Comparing scales, current: %v, desired: %v", s.currentScale, s.desiredScale)
+	s.desiredScale = nMinMax
 
 	if s.desiredScale > s.currentScale {
 		// Scale up immediately.
-		go s.scaleFunc(s.desiredScale, false)
-		s.scaleDownStarted = false
+		log.Printf("Scaling up immediately")
+		go func() {
+			if err := s.scaleFunc(s.desiredScale, false); err != nil {
```
nstogner, line 83: Idiomatic way of doing this is to only check error condition, and return early on error instead of `if ... else`

pkg/deployments/manager.go
samos123: done, great catch!

pkg/deployments/scaler.go
samos123: Thanks for taking the time to do a deeper analysis on how it differs. I agree that setting it to zero value is cleaner and have fixed it.

@nstogner regarding your implementation, it's important that we always check and log the error of s.scaleFunc because otherwise it will be hard to debug if there is a scaling issue. Was there any reason we weren't originally doing error checks? and any reason why your implementation does not? Just making sure I'm not doing something useless with the error checking.

pkg/deployments/scaler.go
samos123: but s.applyMinMax will call `Lock()` so that's why I did this. Is that not needed even when it also calls `Lock()`?

samos123: I'm pretty sure this would result in a deadlock unless I use the local var, see code example here which simulates what would happen if I were to call s.desiredCale = s.applyMinMax(n) within the Lock()
```

func main() {
	mtx := sync.Mutex{}
	mtx.Lock()
	fmt.Println("Locked")
	mtx.Lock()
	fmt.Println("Double locked")
}
```

pkg/deployments/scaler.go
samos123: fixed, agree that does seem a bit cleaner

pkg/deployments/scaler.go
samos123: ok will do. I will do 2 PRs and then base 2nd PR on the 1st refactor PR that doesn't change logic

pkg/deployments/scaler.go
samos123: I think just not initializing does the same thing? Can you verify the latest code? it seems to all work as expected right now.

pkg/deployments/manager.go
samos123: so we can do HasModel because stats aggregates stats of all deployments and autoscaler is active on all deployments. I had to do this to get it to work.

pkg/autoscaler/autoscaler.go
samos123: Lingo right now has autoscaler active on all deployments because ActiceRequests returns the active requests for all deployments. This if statement can be removed once we fix https://github.com/substratusai/lingo/issues/59 and use Pod endpoints and only add endpoints of Deployments that have a lingo model annotation.

samos123: To answer your question, this happens in all scenarios.

pkg/autoscaler/autoscaler.go
```diff
@@ -93,6 +93,11 @@ func (a *Autoscaler) Start() {
...
+				log.Printf("Deployment: %v has no model annotations, skipping", deploymentName)
```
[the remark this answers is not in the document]
samos123: I think we can remove this log once we remove this if statement. The logs are confusing without a log saying that the deployment is ignored. The reason being that earlier in the log it shows aggregating stats of e.g. kubernetes which might make people think it's active on the kubernetes service.

pkg/deployments/scaler.go
nstogner: You can call it before the `Lock()`

samos123: but I shouldn't modify `s.desiredScale` before calling `Lock()`?

samos123: after quick chat, I fixed this by removing the lock in applyMinMax and moving the call to applyMinMax into the lock itself.

pkg/deployments/scaler_test.go
```diff
@@ -89,3 +76,111 @@ func TestSetDesiredScale(t *testing.T) {
...
+	mockScaleMtx.Unlock()
+	s.UpdateState(7, 1, 10)
+	// Wait for scale down delay to pass
+	time.Sleep(2*time.Second + 10*time.Millisecond)
+
+	// Trigger a scale down
+	s.SetDesiredScale(3)
+	time.Sleep(1 * time.Second)
+	mockScaleMtx.Lock() // Ensure consistency of the checked state
+	if scaleFuncCalled != false {
```
nstogner, line 173: You could also use `require.True(t, scaleFuncCalled)` - https://pkg.go.dev/github.com/stretchr/testify/require#True

nstogner (approved):
LGTM - added one small suggestion in tests

Issue #59, opened 2024-01-13
Service name shouldn't have to match deployment name


samos123:
Currently it's hardcoded that a service name should match deployment name https://github.com/substratusai/lingo/blob/8c381444dfd2a957964f471ae9cb528cb5b666fd/pkg/endpoints/manager.go#L93

See relevant code in endpoints/manager.go:
```
	const serviceNameLabel = "kubernetes.io/service-name"
	serviceName, ok := slice.GetLabels()[serviceNameLabel]
	if !ok {
		log.Printf("no service label on endpointslice: %v", req.Name)
		return ctrl.Result{}, nil
	}

	if priorLen != len(ips) {
		// TODO: Currently Service name needs to match Deployment name, however
		// this shouldn't be the case. We should be able to reference deployment
		// replicas by something else.
		r.EndpointSizeCallback(serviceName, len(ips))
	}
```

We should try and get rid of this hardcoded assumption that service name is same as deployment name.

Some ideas:
* Utilize a pod watcher and get IP addresses directly from the pod resource, this would allow us to get rid of the requirement of defining any Service.
* Have Lingo work at the K8s Service level instead of having annotations at the Deployment resource

samos123:
@alpe what are your thoughts on this? I personally prefer the idea of having a Pod reconciler.

alpe:
🤔 The pods do not contain the `lingo.substratus.ai/models` annotation but the deployment does, currently.
When watching the pods, you would have to do additional queries to follow the path to the model:`pod.ownerReferences ->replicaSet.ownerReference->deployment.annotations`. 
Moving the model annotation to the pods spec can be a short cut. But as pods are scaled up and down, we would still have to watch the deployments as pods may be scaled down to 0 and not be found on lingo bootstrap.

With the annotation on the service, we don't have the state problem as the service does either exist or not. But we have a hard time to fetch the replica set to scale up/ down. There is no direct path. So either we rely on the convention or duplicate the model data which seem error prune.

The autoscaler does also use the endpoint manager to fetch all lingo instances to collect metrics. Any solution needs to take care of this as well.

nstogner:
Note: Drawback to watching Pods is that it makes Lingo less scalable / more resource intensive at larger scales. Pods are noisy objects that can generate a lot of api-server traffic for the controller to watch and cache. EndpointSlices change much less.

```bash
k get endpointslices -w
NAME             ADDRESSTYPE   PORTS     ENDPOINTS    AGE
[...]

PR #80, opened 2024-02-14
Spike: Reconcile pods for model ip addresses

alpe:
🚧 Spike to discuss a solution for #59

### DIFF
--- a/pkg/endpoints/manager.go
+++ b/pkg/endpoints/manager.go
@@ -4,19 +4,22 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"strings"
 	"sync"
 
+	corev1 "k8s.io/api/core/v1"
 	disv1 "k8s.io/api/discovery/v1"
-
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 func NewManager(mgr ctrl.Manager) (*Manager, error) {
-	r := &Manager{}
-	r.Client = mgr.GetClient()
-	r.endpoints = map[string]*endpointGroup{}
-	r.ExcludePods = map[string]struct{}{}
+	r := &Manager{
+		svcEndpoints: make(map[string]*endpointGroup),
+		podEndpoints: make(map[string]*endpointGroup),
+		ExcludePods:  make(map[string]struct{}),
+		Client:       mgr.GetClient(),
+	}
 	if err := r.SetupWithManager(mgr); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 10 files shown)


pkg/proxy/handler.go
```diff
@@ -85,7 +85,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	log.Println("Waiting for IPs", id)
-	host, err := h.Endpoints.AwaitHostAddress(r.Context(), deploy, "http")
+	host, err := h.Endpoints.AwaitHostAddress(r.Context(), modelName, "http")
```
alpe, line 88: Please note that IPs are stored by model now and not by deployment as before.

pkg/endpoints/manager.go
```diff
@@ -28,19 +31,75 @@ type Manager struct {
...
+	}
+	switch pod.Status.Phase {
+	case corev1.PodPending:
+		return ctrl.Result{}, nil
+	case corev1.PodRunning:
+		// new or updated instance
+		ports := make(map[string]int32)
+		for _, container := range pod.Spec.Containers {
+			for _, port := range container.Ports {
+				ports[port.Name] = port.ContainerPort
```
samos123, line 70: Question to myself, in K8s can you call the port of a container without it being defined as a ContainerPort?

samos123: I just tested this, seems like you can.  So purely relying on pods to get the port might not be correct either.

Here is the test I ran:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-nginx
spec:
  selector:
    matchLabels:
      run: my-nginx
  replicas: 2
  template:
    metadata:
      labels:
        run: my-nginx
    spec:
      containers:
      - name: my-nginx
        image: nginx
```

Create a svc:
```
  ports:
  - port: 80
    protocol: TCP
    targetPort: 80
  selector:
    run: my-nginx
```

Check the endpoints:
```
my-nginx     10.244.0.5:80,10.244.0.6:80   2m49s
```

Notice that endpoint does have the correct port but pod information does not:
```
  containers:                                           
  - image: nginx                                        
    imagePullPolicy: Always                             
    name: my-nginx      
    resources: {}                                                                                               
    terminationMessagePath: /dev/termination-log        
    terminationMessagePolicy: File                                                                              
    volumeMounts:                                       
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: kube-api-access-dsn4b                       
      readOnly: true 
```

alpe: Good catch. Not sure what is the best approach in this scenario. I think it would be fair to log an error and ignore the pod but I am curious for other options
Btw, with the pods as source for IPs, we don't need the service anymore.

samos123:

PR #81, opened 2024-02-15
GitHub action for creating release

samos123:


nstogner (commented):
How is this different from the create-release button?

samos123:
So all I need to do is `git tag v0.1.0 && git push` instead of doing that and then hoping I remember to do clicky clicky button.

samos123:
@nstogner lets get this merged so our github project shows having releases with change logs whenever i push a new tag. I tested this out in vllm-docker project and it's very convenient.

samos123:
Merging, can revert if you have a strong opinion against it.

PR #88, opened 2024-03-30
Messaging integration (GCP PubSub, AWS SQS, Kafka, etc)

### LINKED ISSUE #86: Batch support through Pub/Sub
Ability for Lingo to directly listen and publish to a pub/sub topic

Messages should be using JSON

Request Message format:
```
{
  metadata: dict # required, used to store additional info about the request and returned in the response
  body: dict # required, openai API compatible HTTP body
  options: dict # optional
}
```

Output format:
```
{ 
  metadata: dict 
  response: {"status_code": 200, "body": '{"choices": [{"index": 0, "text": "completion of prompt"}], ..}'
}
```

The output message also contains the metadata of the request. This is needed such that end-users can easily join back the original request with the response they receive in another pubsub topic. Metadata is required because otherwise users risk sending requests but having no way to correlate the request and response.

MV error handling:
* Consider all errors as equal and retry for max of `x` retries no matter what kind of error it is. It may be a waste to retry a malformed request 3 times, but we can optimize later, not a big deal for MVP imo.
* The following behavior for any kind of errors:
  * After x retries, send back a response using the same format `{"metadata": dict, "response": dict}`
  * In case it's valid HTTP response with a body and a status code, include the status code and body in the response dict  
  * In case we were not able to get a valid HTTP response at all, set status_code to -1 and set the field "error" in the response dict where the text string is a helpful actionable text string for the end-user.

Things to consider:
* min and max concurrent requests per lingo instance and being able to configure that to maximize utilization of resources prevent spikey scale ups and scale downs
* Multi-region support by having lingo instances in multiple regions listen to the same pub/sub topic
* How to handle different kind of errors with retries

nstogner:
Add messaging integration (consume requests and produce responses via a messaging system).

Implemented via [gocloud](https://gocloud.dev/) package to allow for future cross-cloud support.

Also refactors configuration to use environment variables exclusively.

Fixes #86


samos123:
I was thinking to keep it simple initially it might be easier to listen to the topic and do a curl to localhost? That way we can re-use all the existing logic instead of implementing it twice. We can optimize later.

nstogner:
We need a way of determining how many concurrent messages to process and when to stop pulling new messages of off the subscription. The easiest way is to invoke the functions directly b/c thats where we have the info. The alternative would be to always process a given number of requests concurrently and wait for the proxy handler to return. I see that as being harder to debug and also it adds another layer of concurrency settings.

nstogner:
NOTE: Currently the code is expecting a request message that looks like the following which is slightly different from the issue's description.

```json
{
  # Standard OpenAI fields
  "model": "...",
  "prompt": "What is the ...",

  # Lingo-specific subscriber fields
  "path": "/v1/completions",
  "metadata": {
    "optional-key": "optional-val"
  }
}
```

I am back-n-forth on whether we should nest all of the OpenAI fields under `.body` or not. To avoid collisions with future OpenAI APIs it might be a good idea. It also might be a good idea so that users of languages with well defined types can use OpenAI native types for the `.body` field.

I am planning on making this update soon. @samos123, thoughts?

nstogner:
A few changes are still needed:

* Add subscriber instantiation to `main.go`.
* Add concurrent message handling.
* Test against a live pubsub topic.

samos123:
I do think it's the best longer term approach so we can have more control over queueing. It's important that we only ack messages that have their response sent back to a pubsub topic. So we would still have to wait. I think there might be a timeout from pubsub by when it needa an ack

nstogner:
Appears to be a race somewhere in the integration tests:

```
    util_test.go:67: 
                Error Trace:    /Users/nick/work/substratusai/lingo.pubsub/tests/integration/util_test.go:71
                                                        /usr/local/Cellar/go/1.22.0/libexec/src/runtime/asm_amd64.s:1695
                Error:          Not equal: 
                                expected: 1
                                actual  : 0
                Messages:       scale-up should have occurred
    util_test.go:67: 
                Error Trace:    /Users/nick/work/substratusai/lingo.pubsub/tests/integration/util_test.go:67
                                                        /Users/nick/work/substratusai/lingo.pubsub/tests/integration/subscriber_test.go:44
                Error:          Condition never satisfied
                Test:           TestSubscriber
                Messages:       waiting for the deployment to be scaled up
--- FAIL: TestSubscriber (8.01s)
FAIL
```

nstogner:
Most recent commit appears to have fixed the race condition in the integration tests - running a lot of back-to-back tests to make sure now.

nstogner:
Ready for testing on GCP.

A very rudimentary test shows a request and response.

Running the controller:

```bash
MESSENGER_URLS='gcppubsub://projects/my-project/subscriptions/lingo-requests-sub|gcppubsub://projects/my-project/topics/lingo-responses' go run ./cmd/lingo/main.go
```

Sending a request:

```bash
$ gcloud pubsub topics publish lingo-requests \
  --message='{"path":"/v1/completions", "metadata":{"a":"b"}, "body": {"model": "mdl-1"}}'
messageIds:
- '10824071783903012'
```

I get a response:

```bash
$ gcloud pubsub subscriptions pull lingo-responses-sub --auto-ack
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────┬───────────────────┬──────────────┬──────────────────────────────────────┬──────────────────┬────────────┐
│                                                    DATA                                                    │     MESSAGE_ID    │ ORDERING_KEY │              ATTRIBUTES              │ DELIVERY_ATTEMPT │ ACK_STATUS │
├────────────────────────────────────────────────────────────────────────────────────────────────────────────┼───────────────────┼──────────────┼──────────────────────────────────────┼──────────────────┼────────────┤
│ {"metadata":{"a":"b"},"status_code":404,"body":{"error":{"message":"backend not found for model: mdl-1"}}} │ 10824059966496759 │              │ request_message_id=10824071783903012 │                  │ SUCCESS    │
└────────────────────────────────────────────────────────────────────────────────────────────────────────────┴───────────────────┴───────────
```

tests/integration/messenger_test.go
```diff
@@ -0,0 +1,115 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	t.Logf("Waiting for response error message: id %q", id)
+	resp, err := testResponsesSubscription.Receive(ctx)
+	require.NoError(t, err)
+	resp.Ack()
+
+	require.JSONEq(t, fmt.Sprintf(`
+{
```
nstogner, line 68: Note: error format - `.body` should match OpenAI's errors.

pkg/messenger/messager.go
```diff
@@ -0,0 +1,317 @@
...
+			return err
+		}
+
+		log.Println("Received message:", msg.LoggableID)
+
+		m.handleRequest(context.Background(), msg)
+
+		// Slow down a bit to avoid churning through messages and running
+		// up cloud costs when no meaningful work is being done.
+		if consecutiveErrors := m.getConsecutiveErrors(); consecutiveErrors > 0 {
```
samos123, line 80: I think there is also a risk that occasionally there is a short spike of errors that would slow things down. I think

cmd/lingo/main.go
```diff
@@ -85,18 +61,77 @@ func run() error {
...
+		ScaleDownDelay int `env:"SCALE_DOWN_DELAY, default=30"`
+		BackendRetries int `env:"BACKEND_RETRIES, default=1"`
+
+		// MessengerURLs is a list of (comma-separated) URLs to listen for requests and send responses on.
+		//
+		// Format: <request-subscription1>|<response-topic1>,<request-subscription2>|<response-topic2>,...
+		//
+		// URL Examples:
+		//
+		// Google PubSub:
```
samos123, line 79: I would like to see an example in this format
"gcppubsub://projects/my-project/subscriptions/my-subscription|gcppubsub://projects/myproject/topics/mytopic"

samos123:
Very nice! I think we should get this merged in as an experimental MVP and iterate over it. I fixed the docker build and e2e tests by upgrading our Docker image to golang 2.22.

I've also verified this works in a GCP environment that has mistral and mixtral deployed.

samos123 (approved):

samos123:
One thing I don't get is how do we limit the maximum amount of concurrent open requests? Seems right now there is no way to set such limit?

nstogner:
> One thing I don't get is how do we limit the maximum amount of concurrent open requests? Seems right now there is no way to set such limit?

The call to `.EnqueueAndWait` should do this: https://github.com/substratusai/lingo/pull/88/files#r1546379421

pkg/messenger/messager.go
nstogner: True, we will probably need to tune this over time. I think its a good thing to slow stuff down when errors start building up. Right now the wait time will go back to zero once a single message is processed successfully.

This delay account for spontaneous failures that might creep up overnight, but also is intended to mitigate a failure that might happen if some job send a million malformed requests into a topic.

cmd/lingo/main.go
nstogner: Agreed

pkg/messenger/messager.go
nstogner: Added comment containing these thoughts

cmd/lingo/main.go
nstogner: Done

nstogner:
Added Issue to track retry functionality: https://github.com/substratusai/lingo/issues/89

I am good to merge as-is.

PR #91, opened 2024-04-06
messenger: improve concurrent handling

samos123:
Messenger will now default to a 1000 concurrent handlers. You can optionally specify the concurrency handlers for each URL pair, e.g. like this: `gcppubsub://projects/my-project/subscriptions/my-subscription|gcppubsub://projects/myproject/topics/mytopic|1000`

The concurrency is needed because otherwise the EnqueueAndWait will block until scale up happens. This makes sure that the messenger logic reflects how HTTP logic behaves as well.

### DIFF
--- a/cmd/lingo/main.go
+++ b/cmd/lingo/main.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -94,18 +95,29 @@ func run() error {
 	}
 
 	type messengerURLPair struct {
-		requests  string
-		responses string
+		requests    string
+		responses   string
+		maxHandlers int
 	}
 	var messengerURLPairs []messengerURLPair
 	for _, s := range cfg.MessengerURLs {
 		parts := strings.Split(s, "|")
-		if len(parts) != 2 {
+		if len(parts) != 2 && len(parts) != 3 {
 			return fmt.Errorf("invalid subscription URL: %q", s)
 		}
+		var maxHandlers int = 1000
+		var err error
+		if len(parts) == 3 {
+			maxHandlers, err = strconv.Atoi(parts[2])
+			if err != nil {
+				return fmt.Errorf("error converting messenger URL maxHandlers string %v to int: %w",
+					parts[2], err)
+			}
+		}
 		messengerURLPairs = append(messengerURLPairs, messengerURLPair{
-			requests:  parts[0],
-			responses: parts[1],
+			requests:    parts[0],
+			responses:   parts[1],
+			maxHandlers: maxHandlers,
 		})
 	}
 
@@ -196,6 +208,7 @@ func run() error {
 			ctx,
 			msgURL.requests,
 			msgURL.responses,
+			msgURL.maxHandlers,
 			deploymentManager,
 			endpointManager,
 			queueManager,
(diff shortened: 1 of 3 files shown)


cmd/lingo/main.go
```diff
@@ -73,10 +74,14 @@ func run() error {
...
 		//
 		// Format: <request-subscription1>|<response-topic1>,<request-subscription2>|<response-topic2>,...
+		// You can optionally also specify the max number of handlers for each messenger URL pair.
+		// If not specified, the default is 1000.
+		// Format with setting max concurrency: <request-subscription1>|<response-topic1>|<max-handlers1>,
 		//
 		// Examples:
 		//
 		// Google PubSub:		"gcppubsub://projects/my-project/subscriptions/my-subscription|gcppubsub://projects/myproject/topics/mytopic"
+		// with maxHanlders:	"gcppubsub://projects/my-project/subscriptions/my-subscription|gcppubsub://projects/myproject/topics/mytopic|1000"
```
nstogner, line 84: `maxHanlders` spelling

cmd/lingo/main.go
```diff
@@ -73,10 +74,14 @@ func run() error {
 		// MessengerURLs is a list of (comma-separated) URLs to listen for requests and send responses on.
 		//
 		// Format: <request-subscription1>|<response-topic1>,<request-subscription2>|<response-topic2>,...
+		// You can optionally also specify the max number of handlers for each messenger URL pair.
+		// If not specified, the default is 1000.
+		// Format with setting max concurrency: <request-subscription1>|<response-topic1>|<max-handlers1>,
```
nstogner, line 79: Slight opinion: In my head, the `|` character was used to symbolize a linux pipe. It is a little confusing to tack on options with additional pipes, what about other settings? Will they be added with more `|`'s? I would prefer something more like: `<request-sub>|<response-topic>,concurrency:1000`. Longer term I think we might need to move to a yaml/json config file.

pkg/messenger/messager.go
```diff
@@ -73,31 +75,28 @@ func (m *Messenger) Start(ctx context.Context) error {
 
 		log.Println("Received message:", msg.LoggableID)
 
-		m.handleRequest(context.Background(), msg)
-
-		// Slow down a bit to avoid churning through messages and running
```
nstogner, line 78: I don't think we should remove the error logic in this PR. I think we really don't want to be churning through 100 million requests, churning out 100 million error responses, leaving the user with a hefty bill. Too many things can go wrong and lead to that situation IMO.

nstogner (changes requested):
The concurrency logic looks good but the removal of the error handling backoff makes me very nervous.

cmd/lingo/main.go
nstogner: Another option would be to specify a global value in the controller that applies for the entire controller. It could either be implemented as `CONCURRENCY_PER_MESSENGER` or `MESSENGER_TOTAL_CONCURRENCY`

samos123: I was wondering why the `|` and did not think about it as a pipe but that does make sense! My mind is very context based. I only think in pipes when I'm in my shell.

I prefer to keep as-is for now since the default of 1000 is probably good enough for most people.

pkg/messenger/messager.go
samos123: fixed, brought it back

nstogner (approved):
LGTM!

PR #94, opened 2024-04-09
Fix missing return - causing duplicate calls to sendResponse - triggering a panic

### LINKED ISSUE #93: Messenger: Panic panic: Ack/Nack called twice on
There seems an unhandled error which causes lingo to crash and then the pod will get started, directly causing in termination of all pods because lingo will scale down from e.g. 100 replicas directly back to 1 after a restart.

This is the error seen in the logs before the failure:
```
panic: Ack/Nack called twice on message: &{LoggableID:10628088997508317 Body:[...
```

nstogner:
Fixes #93


samos123 (approved):
Ideally we have a test case that fails and is fixed by this PR as part of the PR, but leaning towards speed and lets merge first.

PR #97, opened 2024-04-13
Add hostname to message metadata

nstogner:
Indicates which Pod processed a request


samos123 (commented):
Can we default to an automatically detected hostname?

PR #107, opened 2024-07-05
fix #106 improve scale down behavior

### LINKED ISSUE #106: Improve scaling behavior when there are requests waiting to be queued
When concurrency per replica is 100 and average is 100 then it should not try to scale down to from 28 to 1.

See this state:
```
2024/07/05 07:31:07 Average for deployment: llama-3-8b-instruct-vllm: 102.2 (ceil: 1), current wait count: 90
2024/07/05 07:31:07 SetDesiredScale(1), current: 29, min: 1, max: 500
```
There are currently 29 replicas that are each on average processing 102 requests however, lingo in this case decides to scale back to 1 replica.

I think in this case, the logic could be improved that if wait count is higher than 50% of concurrency per replica, it should not try to scale down anything at all.

samos123:
Only scale down by 1 instead of scaling down to 1 when all replicas are actively processing requests.

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -100,7 +100,16 @@ func (a *Autoscaler) Start() {
 			normalized := flt / float64(a.ConcurrencyPerReplica)
 			ceil := math.Ceil(normalized)
 			log.Printf("Average for deployment: %s: %v (ceil: %v), current wait count: %v", deploymentName, flt, ceil, waitCount)
-			a.Deployments.SetDesiredScale(deploymentName, int32(ceil))
+			current := a.Deployments.GetCurrentReplicas(deploymentName)
+			// Scale up based on average wait count.
+			if current < int32(ceil) {
+				a.Deployments.SetDesiredScale(deploymentName, int32(ceil))
+			} else if int(flt) < a.ConcurrencyPerReplica {
+				// Scale down if the average is less than the current scale.
+				a.Deployments.SetDesiredScale(deploymentName, current-1)
+			} else {
+				log.Printf("No change for deployment: %s, current: %v, desired: %v", deploymentName, current, ceil)
+			}
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


nstogner:
If all replicas are actively processing requests my understanding is that any active request should be considered in the accounting of in-progress requests (contribute to the average that drives the scale).

I think this PR would effectively just rate limit scale downs to 1 every 3 seconds (the period for running the autoscaling loop).

I think a better approach might be to adjust the moving average to encompass a longer time window. Time window should probably be configurable. See:
https://github.com/substratusai/lingo/blob/55f112224aae5fea090fa13b3bc06e006813b9a5/cmd/lingo/main.go#L188
I will start on a PR for this.

nstogner:
Another possible cause: If this is in conjunction with using a pubsub integration, it is possible that this flapping is due to error backoff: https://github.com/substratusai/lingo/blob/55f112224aae5fea090fa13b3bc06e006813b9a5/pkg/messenger/messager.go#L105

I think this could cause it b/c we don't know the total in-queue messages, we drive the moving average off of how many we pull from the queue over time. I will add an issue to track making this backoff configurable as well.

samos123:
What would you expect this code to return in this scenario?
```
		for deploymentName, waitCount := range stats.ActiveRequests {
			log.Println("Is leader, autoscaling")
			avg := a.getMovingAvgQueueSize(deploymentName)
			avg.Next(float64(waitCount))
			flt := avg.Calculate()
                        normalized := flt / float64(a.ConcurrencyPerReplica)
                        ceil := math.Ceil(normalized)
                        a.Deployments.SetDesiredScale(deploymentName, int32(ceil))

```

concurrency per replica = 100
replicas = 10
each replica is constantly processing 90 items
waitCount = ~0-1

My expectation is that setDesired scale should be 9 or 10.
However the way it's implemented today it would result in setDesired scale =1 and here is why:
avg = ~90
flt= something like 88
ceil=1

nstogner:
`each replica is constantly processing 90 items` --> this would mean that there are `900` total active requests, so I would think:

```
flt = 900 (10 replicas processing 90 requests each)
concurrencyPerReplica = 100
desiredScale = 900 / 100 = 9 
```

nstogner:
Note: if you know the timestamp, it looks like we log this info:
```go
log.Printf("Average for deployment: %s: %v (ceil: %v), current wait count: %v", deploymentName, flt, ceil, waitCount)
```

PR #116, opened 2024-07-09
Recreate subscription on receive message error and add pubsub e2e test

samos123:


nstogner (commented):
Will probably need to test this - might be able to simulate a receive error in the tests - will look

nstogner (approved):

PR #130, opened 2024-08-25
Add nomic embedding model and add model feature filtering

nstogner:
Add Nomic Embedding model:

> nomic-embed-text is a large context length text encoder that surpasses OpenAI text-embedding-ada-002 and text-embedding-3-small performance on short and long context tasks.

 See: https://ollama.com/library/nomic-embed-text

Also: Add query string filtering in the model listing endpoint to avoid OpenWebUI adding embedding models in the dropdown list of chat-able models.

```go
// Example (default):  /v1/models
// Example (single):   /v1/models?feature=TextEmbedding
// Example (multiple): /v1/models?feature=TextGeneration&feature=TextEmbedding
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ models:
       minReplicas: 1
     qwen2-500m-cpu:
       enabled: true
-    nomic-embed-text-cpu
+    nomic-embed-text-cpu:
       enabled: true
 EOF
 


samos123 (approved):

internal/modelcontroller/model_controller.go
```diff
@@ -86,6 +86,14 @@ func (r *ModelReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
 			return ctrl.Result{}, nil
 		}
+		if changed {
+			log.Info("applied resource profile")
+			shouldUpdate = true
+		}
+	}
+	{
+		// Apply self labels based on features so that we can easily filter models.
+		changed := r.applySelfLabels(model)
```
samos123, line 96: nit: you could combine this in single statement
```
if changed := r.applySelfLabels(model); changed {
   shouldUpdate = true
}
```

PR #135, opened 2024-08-27
Add ability to use existing secret for huggingface token

### LINKED ISSUE #133: Support using existing secret containing Huggingface token
I already have a secret in my cluster that contains my Huggingface token. I would like the ability to set the values for the secret name and the key instead of providing the token in my values.yaml 
https://github.com/substratusai/kubeai/blob/8cbfcc99e2246312a79f8e424c8dbd6d424280f3/charts/kubeai/values.yaml#L5-L7

BOsterbuhr:
I also looked at adding the ability to specify the secret key as well but decided to settle on adding the comment in the values.yaml

resolves #133

### DIFF
--- a/charts/kubeai/templates/_helpers.tpl
+++ b/charts/kubeai/templates/_helpers.tpl
@@ -60,3 +60,14 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Create the name of the huggingface secret to use
+*/}}
+{{- define "kubeai.huggingfaceSecretName" -}}
+{{- if .Values.secrets.huggingface.secretName -}}
+huggingface: {{ .Values.secrets.huggingface.secretName }}
+{{- else }}
+huggingface: {{ include "kubeai.fullname" . }}-huggingface
+{{- end }}
+{{- end }}
--- a/charts/kubeai/templates/_helpers.tpl
+++ b/charts/kubeai/templates/_helpers.tpl
@@ -65,9 +65,12 @@ Create the name of the service account to use
 Create the name of the huggingface secret to use
 */}}
 {{- define "kubeai.huggingfaceSecretName" -}}
-{{- if .Values.secrets.huggingface.secretName -}}
-huggingface: {{ .Values.secrets.huggingface.secretName }}
+{{- if .Values.secrets.huggingface.create -}}
+{{- default (include "kubeai.fullname" .) .Values.secrets.huggingface.name }}
 {{- else }}
-huggingface: {{ include "kubeai.fullname" . }}-huggingface
+{{- if not .Values.secrets.huggingface.name -}}
+{{ fail "if secrets.huggingface.create is false, secrets.huggingface.name is required" }}
+{{- end }}
+{{- .Values.secrets.huggingface.name }}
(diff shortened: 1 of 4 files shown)


charts/kubeai/templates/_helpers.tpl
```diff
@@ -60,3 +60,14 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Create the name of the huggingface secret to use
+*/}}
+{{- define "kubeai.huggingfaceSecretName" -}}
+{{- if .Values.secrets.huggingface.secretName -}}
+huggingface: {{ .Values.secrets.huggingface.secretName }}
```
nstogner, line 69: Its a little confusing to me to have huggingfaceSecretName be a key/val pair. Can we change this helper to return a string?

charts/kubeai/templates/secret.yaml
```diff
@@ -1,8 +1,10 @@
+{{- if .Values.secrets.huggingface.token -}}
```
nstogner, line 1: See comment below on using `.create` bool

charts/kubeai/values.yaml
```diff
@@ -5,6 +5,9 @@
 secrets:
   huggingface:
     token: ""
+    # secretName is used to pass the huggingface token value via an existing k8s secret
+    # the value is pulled from the key, "token".
+    secretName: ""
```
nstogner, line 10: To keep things consistent let's follow the same convention of using `.create` and `.name` as is used for `.serviceAccount` lower down in the values file.

(This convention was templated by `helm create`)

```yaml
serviceAccount:
  # Specifies whether a service account should be created
  create: true
  # Automatically mount a ServiceAccount's API credentials?
  automount: true
  # Annotations to add to the service account
  annotations: {}
  # The name of the service account to use.
  # If not set and create is true, a name is generated using the fullname template
  name: ""
```

nstogner (changes requested):
Thanks for the PR! Requesting a few changes to keep things as consistent as possible in the chart.

samos123 (approved):

charts/kubeai/values.yaml
```diff
@@ -4,7 +4,12 @@
 
 secrets:
   huggingface:
+    create: true
```
samos123, line 7: we should probably default to false, so it's not a hard requirement to supply a HF token? @nstogner your thought?

BOsterbuhr: With [the conditionals](https://github.com/substratusai/kubeai/pull/135/files#diff-88a0eb93bdcb1a4e615ab6e8751c47beaeadbbf57e9646ba071dd8bd418da6c7R69-R73) that I added, a HF token is a hard requirement.
Currently if you don't specify a token a secret is still created just with an empty value. When you go to pull a model that requires a token it will fail but you have to read the model's pod logs to figure out why.

nstogner (approved):
LGTM!

nstogner (changes requested):
Lets remove `{{ fail "secrets.huggingface.token must be set if secrets.huggingface.create is true" }}` to avoid requiring a token (will break quickstart)

samos123 (approved):

PR #137, opened 2024-08-27
Publish KubeAI Arm container image

### LINKED ISSUE #134: Build and Publish KubeAI images for ARM64
Currently `substratusai/kubeai` is only available for linux/amd64 on Docker Hub, publishing images for linux/arm64 would be beneficial as ARM nodes are cheaper on AWS than the AMD equivalent.

samos123:
fixes #134

Note the vLLM CPU image does not support Arm, but this commit publishes kubeai compatible arm image. The upstream vLLM GPU image does seem to support Arm out of the box. The Ollama image also supports Arm already.

### DIFF
--- a/.github/workflows/build-push.yml
+++ b/.github/workflows/build-push.yml
@@ -26,6 +26,8 @@ jobs:
         uses: actions/checkout@v4
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx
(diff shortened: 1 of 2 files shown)


samos123:
Getting some weird errors on the vLLM CPU image when building it: https://github.com/substratusai/vllm-docker/actions/runs/10587101588/job/29337454516?pr=28#step:8:548

nstogner (approved):

Issue #132, opened 2024-08-27
Support dynamic LoRA serving


nstogner:
KubeAI should be able to serve dynamically loaded LoRA adapters. Eventually KubeAI could produce these adapters through supporting a finetuning API endpoint, however that can be implemented separately.

I see 2 primary options:

### Option A

KubeAI handles shipping adapters to different server Pods (i.e. `kubectl cp` or shared filesystem) and keeping track of which Pods have which adapters.

### Option B

Server backends handle dynamic loading of adapters themselves and KubeAI just keeps track of what Pods already have adapters in order to load balance effectively.

Not yet supported in vLLM: https://github.com/vllm-project/vllm/issues/6275
Currently supported by Lorax: https://github.com/predibase/lorax

nstogner:
We have a few competing priorities right now:

1. LoRA
2. vLLM Cache-aware routing (should greatly improve perf when model replicas >1)
3. Scaling across infra (i.e. CPU -> GPU -> TPU)

Will work to define some dates soon. Sounds like LoRA would be your number 1 priority out of those?

ffais:
Yes, serve dynamically loaded LoRA adapters is our number 1 priority.

nstogner:
Can you provide some details on your use case so that we can make sure that we will solve it? Where do you store the adapters? What would be the total expected number of adapter variants you would have for a given model? How are you serving them today?

ffais:
At this moment we have an instance of Lorax deployed with different base model and for each one 4/5 adapters. We're storing all adapters on S3.

nstogner:
Started work on this, we will get this tackled as our next big feature.

PR #140, opened 2024-08-28
Model management proposal

### LINKED ISSUE #139: Create a proposal for managing model weights and adapters
Take into account:

* #3 
* #132 
* #138

nstogner:
Fixes #139


proposals/model-management.md
```diff
@@ -0,0 +1,67 @@
...
+  # While standard storage profile names are recommended, they should be open ended
+  # to allow for multiple different admin-defined varieties:
+  MyCustomBucket:
+    bucket:
+    # ...
+```
+
+### Bucket storage
+
+This mode relies on access to a bucket. To support zero external dependencies and local deployment this bucket could be provided by a Helm-deployed Minio (possible License issue?) / [SeaweedFS](https://github.com/seaweedfs/seaweedfs) (mature/secure?) instance.
```
samos123, line 54: Seaweed would be my choice. I looked into Seaweedfs in the past for same use case. Seaweed is lightweight and has a single binary option.

proposals/model-management.md
```diff
@@ -0,0 +1,67 @@
...
+* KubeAI should be able to be run on clouds with heterogenous disk features (limited support for ReadOnlyMany)
+
+## Solution
+
+KubeAI should support configurable storage for both models and adapters. These settings should be configurable at the `kind: Model` level as some variants (like GCP Hyperdisk ML) are expensive and generally only recommended for larger models.
+
+```yaml
+kind: Model
+spec:
+  storageProfile:
```
samos123, line 24: is StorageProfile optional? Can we just leave it out and let it pull from HF?

proposals/model-management.md
```diff
@@ -0,0 +1,67 @@
...
+    bucket:
+      url: s3://...
+      # ...
+  ReadWriteManyDisk:
+    disk:
+      storageClass: aws-efs
+      # ...
+  ReadManyDisk:
+    disk:
+      storageClass: gcp-hyperdisk
```
samos123, line 43: so would this be a single hyperdisk shared with all models? and each model is in a subdirectory?

proposals/model-management.md
nstogner: For pulling models, yes, thinking we have a storage profile option of `None`. I added a block for default storage profiles when none are specified on the Model.

proposals/model-management.md
nstogner: Updated the proposal a little more - ML Hyperdisk would be mounted once to populate it with a single model, then mounted read-only by multiple servers.

proposals/model-management.md
```diff
@@ -0,0 +1,109 @@
...
+    adapters: Bucket
+```
+
+Storage profiles should be defined at the KubeAI config.yaml level:
+
+```yaml
+# Set on Model objects if .spec.storageProfile is undefined.
+defaultStorageProfiles:
+  model: None # None is a special storage mode where each server pulls the model from source.
+  adapters: Bucket
```
samos123, line 38: Maybe adapters should be None too and then it pulls from HF?

I am thinking we have a default where there is no requirement for any buckets, not even locally within the cluster.

nstogner: I was thinking about the situation where we support finetuning (a job that KubeAI manages the produces LoRA adapters). I was thinking we would need a storage mechanism not just a retrieval strategy. However you made me think we could just push those back up to huggingface with the token they already configured.

nstogner: Updated

samos123:
I think I get the proposal now! So would it be possible to express this as well once we support GCS/S3 as a model location?
```
kind: Model
spec:
  url: "gcs://my-bucket/Meta-Llama-3.1-7b-full-training"
```

The use case would be pushing full merged models to a specific location. Or some private model that uses llama 3.1 7b architecture but was trained from scratch.

proposals/model-management.md
```diff
@@ -0,0 +1,149 @@
...
+  - name: sarcasm
+    url: "hf://myrepo/sarcastic-lora-adapter-for-llama-3-8b-Instruct"
+
+# Model that explicitly asks for a cache profile
+kind: Model
+spec:
+  url: "hf://meta-llama/Meta-Llama-3.1-405B-Instruct"
+  cacheProfile: HyperDisk
+
+# A Model without any cache specified will use the default set in KubeAI's config.yaml
```
samos123, line 42: and the default cacheProfile on installation is cacheProfile: None?

proposals/model-management.md
```diff
@@ -0,0 +1,149 @@
...
+
+If a user submits a finetuning request (`POST /v1/fine_tuning/jobs`), they have the option to specify a location to store the adapters:
+
+```json
+# OpenAI-compatible finetune job request
+{
+  # Model.metadata.name
+  "model": "llama-3.1-8b",
+  # KubeAI-Only: Training file could be specified as URL: "s3://..."
+  "training_file": "<id-of-file-uploaded-via-files-api>",
```
samos123, line 64: +1 to using something like s3:// or http://

proposals/model-management.md
```diff
@@ -0,0 +1,149 @@
...
+Model cache profiles and finetune default storage locations should be defined at the KubeAI `config.yaml` level:
+
+```yaml
+finetuning:
+  storage:
+    defaultURLTemplate: "s3://my-bucket/finetuning/{model}/{job}"
+  
+models:
+  cache:
+    defaultProfile: Bucket # Could also be "None"
```
samos123, line 97: if not specified I'm assuming the helm default values would look like this:
```
models:
  cache:
    defaultProfile: ""
    profiles: {}
```

proposals/model-management.md
```diff
@@ -0,0 +1,149 @@
...
+  - Typically provisioned outside of the k8s API
+
+##### Static Volume Storage
+
+This mode relies on a Kubernetes PersistentVolumes for storage.
+
+- How it works: mounted as ReadWriteOnce to initialize, then ReadOnlyMany for serving
+- Examples: GCP Hyperdisk ML
+- NOTEs:
+  - Only supported for models, NOT adapters!
```
samos123, line 142: I don't get why this wouldn't be supported for adapters?

proposals/model-management.md
```diff
@@ -0,0 +1,149 @@
...
+- NOTEs:
+  - Works for models and adapaters.
+  - Typically provisioned outside of the k8s API
+
+##### Static Volume Storage
+
+This mode relies on a Kubernetes PersistentVolumes for storage.
+
+- How it works: mounted as ReadWriteOnce to initialize, then ReadOnlyMany for serving
+- Examples: GCP Hyperdisk ML
```
samos123, line 140: GCP Persistent Disk ReadManyOnly is another option

proposals/model-management.md
nstogner: Maybe more explicit with None:

```yaml
models:
  cache:
    defaultProfile: None
    profiles: {}
```

proposals/model-management.md
nstogner: Disks like Hyperdisk ML (and other PDs in GCP) can be mounted as ReadWriteOnce or ReadOnlyMany ... Supporting this type of disk for adapters is complicated:
* A finetune Job needs to output adapters by mounting a new disk in RWO mode (doable - but forces 1:1 adapter-to-disk which is probably not justified based on adapter's being small in size)
* Model servers need to be able to mount these adapter-specific disks dynamically to swap the adapter that is being used for a given server like vLLM  that is serving a base model - see #132 (I dont know of a way of doing tackling this as volumes specified in a Pod spec are not dynamic - require Pod recreation)

*...Will add these details to proposal*

proposals/model-management.md
nstogner: Good callout - will add that in but mention the max-attachment limitation

proposals/model-management.md
nstogner: yep!

BOsterbuhr: When you say "will use the default set in KubeAI's config.yaml" does that mean the model catalog at `charts/kubeai/charts/models/values.yaml`? If so, and a model is used that isn't in the model catalog, will cacheProfile be a required field?

samos123 (approved):

michaelfeil:
@nstogner Feedback - potentially the similar/samegoes for training data. Training data usually exceeds request size, and duration exceeds timeout. S3 might be a better fit wrt to persistence.

PR #147, opened 2024-08-30
Remove GKE node selector from default L4 profile

### LINKED ISSUE #146: Node Selector should not be specified in default Helm values.yaml file
Node selector for a particular GPU type should only be set when deploying to a specific cloud:

https://github.com/substratusai/kubeai/blob/6c44e8044794d0d1e814018a3f8cdd01c005347f/charts/kubeai/values.yaml#L34

nstogner:
Fixes #146


samos123:
Should we also rename the resource profile from L4 to GPU ?

samos123 (approved):

PR #154, opened 2024-08-30
Map resource profiles to images & add resource profile doc

### LINKED ISSUE #152: Support accelerator type in resource profile
Some bespoke NVIDIA GPU device plugins do not use `nvidia.com/gpu` as the resource name.

In addition we also need to support AMD GPU images.

One potential way of expressing this:
```
resourceProfiles:
  CPU:
    requests:
      cpu: 1
      memory: "2Gi"
  EXOTICL4:
    type: NVIDIAGPU
    limits:
      nvidia.com/gpu-SOME-SILLY-SUFFIX: "1"
  L4:
    type: NVIDIAGPU
    limits:
      nvidia.com/gpu: "1"
    requests:
      nvidia.com/gpu: "1"
      cpu: "6"
      memory: "24Gi"
  MI300:
    type: AMDGPU
    limits:
      nvidia.com/gpu: "1"
    requests:
      nvidia.com/gpu: "1"
      cpu: "6"
      memory: "24Gi"
```

nstogner:
Define imageNames for given model servers, reference imageNames in resourceProfiles.

Switch reference resource profiles to be lowercase to support the possibility of introducing a ResourceProfile CRD in the future (would require a lowercase .metadata.name).

Fixes #152 via a different technique.

### DIFF
--- a/charts/kubeai/charts/models/values.yaml
+++ b/charts/kubeai/charts/models/values.yaml
@@ -9,6 +9,7 @@ catalog:
     features: ["TextEmbedding"]
     owner: intfloat
     url: "hf://intfloat/e5-mistral-7b-instruct"
+    server: VLLM
     resourceProfile: CPU:1
     args:
     - --gpu-memory-utilization=0.9
@@ -18,7 +19,7 @@ catalog:
     features: ["TextGeneration"]
     owner: google
     url: "ollama://gemma2:2b"
-    engine: OLlama
+    server: OLlama
     resourceProfile: CPU:2
   # Llama #
   llama-3.1-8b-instruct-cpu:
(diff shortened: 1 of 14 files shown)


charts/kubeai/charts/models/values.yaml
```diff
@@ -51,29 +51,29 @@ catalog:
...
-    server: VLLM
+    engine: VLLM
     resourceProfile: CPU:1
   opt-125m-l4:
     enabled: false
     features: ["TextGeneration"]
     owner: facebook
     url: "hf://facebook/opt-125m"
-    server: VLLM
+    engine: VLLM
```
samos123, line 69: I would just keep it server for now. I don't think engine is better. It might cause issues for some existing users this rename.

nstogner: Its actually already engine, I switched this commit reverted the change from `engine:` to `server:` that I introduced in the 1st commit of this PR. I agree we should not change this in this PR.

samos123: Lol I see. I guess as long as we keep it as is, then it's good.

samos123:
I had to rebase to resolve conflict with some doc.

samos123:
I fixed some linting issues. I will publish v0.4.3 and update appVersion after merging this PR. Since the new helm chart needs to use updated image.

samos123 (approved):

PR #167, opened 2024-09-02
Add integration tests that ensure quickstart works

samos123:
Simple e2e test written in bash. This ensures that the steps in the README won't be broken by other PRs. More advanced e2e tests can be written in golang. The tests succeed in ~4-5 minutes depending on how fast the download of gemma 2b is.

In some cases the download may take longer than the timeout.


test/quickstart.sh
```diff
@@ -0,0 +1,73 @@
...
+}
+
+# wait for kubeai pod to be ready
+wait_for_pod_ready app.kubernetes.io/name=kubeai
+
+# Ensure the model count is 0
+curl -s -X GET "http://localhost:8000/openai/v1/models" | jq '. | length == 0'
+
+
+helm upgrade --reuse-values --install kubeai charts/kubeai -f - <<EOF
```
nstogner, line 54: This helm upgrade will install the kubeai image that was last published in the chart.

test/quickstart.sh
```diff
@@ -0,0 +1,73 @@
...
+KIND_CLUSTER_NAME=${KIND_CLUSTER_NAME:-kubeai-tests}
+
+if kind get clusters | grep -q ${KIND_CLUSTER_NAME}; then
+  echo "Cluster ${KIND_CLUSTER_NAME} already exists.. reusing it"
+  else
+  kind create cluster --name ${KIND_CLUSTER_NAME}
+fi
+
+# Capture PID and run skaffold devin background
+skaffold run --tail --port-forward &
```
nstogner, line 14: We should make sure this skaffold run never pushes an image or that the image url for skaffold is overwritten with another non-prod repo. I think `skaffold run` can push by default if the target cluster is a remote one. I know this should be ok for kind - but would want to be explicit to safegaurd against this if we switch the cluster someday.

test/quickstart.sh
nstogner: I think what you might want to do in this test is to do a `docker build -t <test-image>` && `kind load docker-image ...` and then override the image name in the chart to be the loaded test image `--set image.repository` / `--set image.tag`.

test/quickstart.sh
nstogner: See comment on helm command below for an alternative approach that might closer test the quickstart as it will more completely rely helm for install.

test/quickstart.sh
samos123: it's working correctly right now based on my testing. Reuse values flag solves this already. I did kubectl describe pod to verify.

Can you give it a try and see if this is really an issue for you?

test/quickstart.sh
samos123: Skaffold doesn't push when it detects kind. However, let me add an explicit push false.

test/quickstart.sh
```diff
@@ -0,0 +1,73 @@
...
+  kubectl wait --for=condition=ready pod -l "$label" --timeout=1200s
+}
+
+# wait for kubeai pod to be ready
+wait_for_pod_ready app.kubernetes.io/name=kubeai
+
+# Ensure the model count is 0
+curl -s -X GET "http://localhost:8000/openai/v1/models" | jq '. | length == 0'
+
+
```
nstogner:
```suggestion
# The previous skaffold step installed KubeAI using the helm chart.
# By using the --reuse-values flag we can just append models to the previous install
# while avoiding overriding the image that skaffold originally built and set in the first
# install.
```

nstogner (approved):
LGTM!

PR #177, opened 2024-09-04
faster whisper implementation

samos123:
The issue with faster whisper is that it does strong validation on the model name. It has to match one of their predefined models. This won't work for KubeAI. So as a workaround, I've removed the model field from the request body before sending it to the backend.

### DIFF
--- a/internal/modelproxy/request.go
+++ b/internal/modelproxy/request.go
@@ -61,58 +61,99 @@ func (pr *proxyRequest) done() {
 // attempts to unmarshal the request body as JSON and extract the
 // .model field.
 func (pr *proxyRequest) parseModel() error {
-	var bodyBuffer bytes.Buffer
-	// Create a TeeReader to read the body content and save it in bodyBuffer.
-	multiReader := io.TeeReader(pr.r.Body, &bodyBuffer)
-
 	// Try to get the model from the header first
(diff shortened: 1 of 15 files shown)


internal/modelproxy/request.go
```diff
@@ -65,21 +67,89 @@ func (pr *proxyRequest) parseModel() error {
 	}
 
 	var err error
+	// TODO (samos123): Improve this to not read the entire body into memory for files.
 	pr.body, err = io.ReadAll(pr.r.Body)
```
nstogner, line 71: I think we should remove the assignment of `pr.body` here for the multipart form and only do this for JSON requests. In our case where we are rewriting the body for multipart forms we duplicating the entire request body in memory. Instead, for multipart form, I think it would be better to parse the form values from the HTTP request, then construct a buffer for what we want to rewrite the body as, and assign the bytes that underly that buffer finally to `pr.body`.

nstogner (changes requested):
Requested an optimization change - if this works as-is feel free to merge and submit an Issue to address the optimization post-merge.

internal/modelproxy/request.go
samos123: Agree, I will tackle that in this PR

nstogner (approved):

PR #191, opened 2024-09-07
Add integration tests, consolidate configuration, bug fixes

### LINKED ISSUE #193: Model fails to scale up to minReplicas after updating minReplicas on existing Model
https://github.com/substratusai/kubeai/pull/191/commits/4b7979f0be07b909b5127e0822562c1e80d2976c

nstogner:
Fixes #193 

* Add integration tests
* Move startup logic to new `command` pkg to reuse within tests
* Consolidate flag config to `config.yaml` with rest of configuration
* Add config validation


charts/kubeai/values.yaml
```diff
@@ -101,10 +101,7 @@ fullnameOverride: "kubeai"
 command:
 - /app/manager
 
-args:
-- --metrics-bind-address=:8080
-- --leader-elect
-- --health-probe-bind-address=:8081
+args: []
```
samos123, line 104: can you separate the helm changes in a separate PR? This will allow us to first publish a new version of the image and then update appVersion and helm changes at the same time. This prevents possibility of incompatibility.

internal/messenger/messenger.go
```diff
@@ -14,12 +14,6 @@ import (
 	"time"
 
 	"gocloud.dev/pubsub"
-	_ "gocloud.dev/pubsub/awssnssqs"
-	_ "gocloud.dev/pubsub/azuresb"
-	_ "gocloud.dev/pubsub/gcppubsub"
```
samos123, line 19: would this break gcp pubsub integration?

samos123: nvm I see this now gets imported in run.go

charts/kubeai/values.yaml
nstogner: Talked about this offline

samos123 (approved):

PR #195, opened 2024-09-08
release helm chart on tag only

samos123:


.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,40 @@
...
+  publish-docs:
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Configure Git Credentials
+        run: |
+          git config user.name github-actions[bot]
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
```
nstogner, line 18: Can you add a comment as to where this email comes from?

nstogner (approved):
LGTM, added comment about comments

.github/workflows/publish-docs.yml
samos123: I have no idea either lol. I think I blindly copy pasted it from mkdocs example. I fixed it and it's now:
```
          git config user.name "$GITHUB_ACTOR"
          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
```
which is the same as the helm release action

PR #198, opened 2024-09-09
fix model endpoint API response format

samos123:
Not sure if this breaks Open WebUI


samos123:
Verified that openwebui still works
image

PR #201, opened 2024-09-11
Add support for container images with built-in models

nstogner:
Adding support for building models into container images. Added concepts doc which explains why one might want to do this (image caching optimization in k8s, air-gapped environments).

* Fix bug where resource profile will over-ride user settings (i.e. `.spec.image` in Model).
* Add docs for embedded models
* Add Dockerfile for embedding model when using `engine: OLlama` Models

TODO:
* Built-in model support for vLLM


docs/how-to/build-models-into-containers.md
```diff
@@ -0,0 +1,43 @@
...
+export MODEL_URL=ollama://qwen2:0.5b
+
+# Customize with your own image repo.
+export IMAGE=us-central1-docker.pkg.dev/substratus-dev/default/ollama-builtin-qwen2-05b:latest
+```
+
+Build and push image. Note: building (downloading base image & model) and pushing (uploading image & model) can take a while depending on the size of the model.
+
+```bash
+# Within kubeai.git...
```
samos123, line 16: I would probably just add a `git clone` command and then remove "This guide assumes you are in the `kubeai.git` repo."

samos123: UX is better if you can just copy paste a command instead of figuring out `kubeai.git` where is that hmm

docs/how-to/build-models-into-containers.md
```diff
@@ -0,0 +1,43 @@
...
+cd ./images/ollama-builtin
+
+docker build --build-arg MODEL_URL=$MODEL_URL -t $IMAGE .
+docker push $IMAGE
+```
+
+Create a model manifest & apply into a cluster with KubeAI installed. NOTE: The only difference between an built-in model image and otherwise is the addition of the `image:` field.
+
+```bash
+cat <<EOF | envsubst > builtin-model.yaml
```
samos123, line 26: Alternative without envsubst
```
kubectl apply -f - << EOF
apiVersion: kubeai.org/v1
kind: Model
metadata:
  name: builtin-model-example
spec:
  features: ["TextGeneration"]
  owner: alibaba
  image: $IMAGE # <-- The image with model built-in
  url: "$MODEL_URL"
  engine: OLlama
  resourceProfile: cpu:1
  minReplicas: 1
  maxReplicas: 3
EOF
```

test/integration/model_profiles_test.go
```diff
@@ -52,3 +53,47 @@ func TestModelProfiles(t *testing.T) {
 		assert.Equal(t, sysCfg().ResourceProfiles[resourceProfileCPU].NodeSelector, pod.Spec.NodeSelector)
 	}, 2*time.Second, time.Second/10, "Resource profile should be applied to the model Pod object")
 }
+
+// TestModelProfiles tests that resource profiles are applied as expected.
+func TestModelResourceProfileWithUserSetValues(t *testing.T) {
```
samos123, line 58: Is there a test case for resourceProfile adds nodeSelector AND user adds nodeSelector in model?

For example resource PRofile adds sam=true as nodeSelector. The user on model sets nick=true

Then the final created model in K8s should have:
```
nodeSelectors:
  sam: true
  nick:true
```

samos123: I see. Let's call this out somewhere.

docs/how-to/build-models-into-containers.md
nstogner: Done

test/integration/model_profiles_test.go
nstogner: Clarified comments in CRD

docs/how-to/build-models-into-containers.md
nstogner: Good suggestion - added

nstogner:
Added issue to track vLLM guide #209

PR #203, opened 2024-09-11
Configurable autoscaling

### LINKED ISSUE #202: Model expose scale down delay
The model should have scale down delay parameter.

```
scaleDownDelay: 300
```

This would prevent a pod from scaling down for 300 seconds after being created.

The default should probably be between 60 and 300 seconds as well since scale ups are slow for larger models.

nstogner:
* Add configurable scale-down-delay, autoscaling interval & window, and target (Fixes #202)
* Add full scale-up-to-down integration test
* Update helm chart values
* Update docs (add autoscaling docs, remove how-to info from concepts and break into separate sections)
* Remove resource-profile-override fields from Model spec


charts/kubeai/charts/crds/crds/kubeai.org_models.yaml
```diff
@@ -964,6 +966,49 @@ spec:
...
+                description: |-
+                  Autoscaling configuration.
+                  Will be set from the AutoscalingProfile if not specified.
+                properties:
+                  disabled:
+                    description: Disabled is used to disable autoscaling.
+                    type: boolean
+                  maxReplicas:
+                    default: 3
+                    description: MinReplicas is the minimum number of Pod replicas
```
samos123, line 979: description is incorrect. It mentions MinReplicas instead of maxReplicas

charts/kubeai/values.yaml
```diff
@@ -60,11 +60,17 @@ resourceProfiles:
   #   nodeSelector:
   #     cloud.google.com/gke-accelerator-type: tpu-v5-lite-podslice
 
+autoscalingProfiles:
+  # NOTE: "defualt" profile is required.
```
samos123:
```suggestion
  # NOTE: "default" profile is required.
```

api/v1/model_types.go
```diff
@@ -63,6 +61,33 @@ type ModelSpec struct {
 
 	// Env variables to be added to the server process.
 	Env map[string]string `json:"env,omitempty"`
+
+	// AutoscalingProfile to be used for autoscaling.
+	AutoscalingProfile string `json:"autoscalingProfile"`
+
+	// Autoscaling configuration.
+	// Will be set from the AutoscalingProfile if not specified.
+	Autoscaling *ModelAutoscaling `json:"autoscaling,omitempty"`
```
samos123, line 70: can i Just set minReplicas=1 and then all other default values would be used?

So for example.
```
autoscalingProfile: "default"
autoscaling:
  minReplicas: 1
```
would result in maxReplicas being set to 100? and target request 100?

samos123 (changes requested):
Looks good except some minor nits and question on behavior.

api/v1/model_types.go
samos123: answering myself, I believe yes due to the defaults configured in the model type itself

api/v1/model_types.go
```diff
@@ -63,6 +61,33 @@ type ModelSpec struct {
...
+	// +kubebuilder:default=100
+	TargetRequests int32 `json:"targetRequests"`
+	// ScaleDownDelay is the minimum time before a deployment is scaled down after
+	// the autoscaling algorithm determines that it should be scaled down.
+	ScaleDownDelay metav1.Duration `json:"scaleDownDelay"`
+	// MinReplicas is the minimum number of Pod replicas that the model can scale down to.
+	// Note: 0 is a valid value.
+	// +kubebuilder:validation:Minimum=0
+	MinReplicas int32 `json:"minReplicas"`
+	// MinReplicas is the minimum number of Pod replicas that the model can scale up to.
```
samos123, line 87:
```suggestion
	// MaxReplicas is the maximum number of Pod replicas that the model can scale up to.
```

internal/config/system.go
```diff
@@ -14,45 +16,98 @@ type System struct {
...
+		s.Autoscaling.Interval.Duration = 10 * time.Second
+	}
+	if s.Autoscaling.TimeWindow.Duration == 0 {
+		s.Autoscaling.TimeWindow.Duration = 10 * time.Minute
+	}
+
+	for name, profile := range s.Autoscaling.Profiles {
+		// NOTE: Keep in sync with defaults in api pkg.
+		if profile.MaxReplicas == 0 {
+			profile.MaxReplicas = 3
```
samos123, line 65:
```suggestion
			profile.MaxReplicas = 100
```

This doesn't match the docs where we say default is 100.

internal/config/system.go
```diff
@@ -14,45 +16,98 @@ type System struct {
...
+
+	if s.Autoscaling.Interval.Duration == 0 {
+		s.Autoscaling.Interval.Duration = 10 * time.Second
+	}
+	if s.Autoscaling.TimeWindow.Duration == 0 {
+		s.Autoscaling.TimeWindow.Duration = 10 * time.Minute
+	}
+
+	for name, profile := range s.Autoscaling.Profiles {
+		// NOTE: Keep in sync with defaults in api pkg.
```
samos123, line 63: I don't understand this comment.

nstogner: Updated with more context

internal/config/system.go
nstogner: Updated to set default value to 3. I am concerned that if we set default max replicas to 100 that teams will only notice it in production. Thoughts?

charts/kubeai/charts/crds/crds/kubeai.org_models.yaml
samos123: probably need to regenerate this CRD?

api/v1/model_types.go
nstogner: `The Model "abc" is invalid: spec.autoscaling.scaleDownDelay: Required value`

The API Server will reject the request b/c if the user sets the autoscaling field they need to supply all required values. Note: the behavior as currently implemented is that if the user sets `.spec.autoscaling` then KubeAI will not override it with values from the profile. With that logic, your case here is confusing and should eventually be mitigated with a validating webhook.

charts/kubeai/charts/crds/crds/kubeai.org_models.yaml
nstogner: Fixed, good catch

docs/how-to/configure-autoscaling.md
```diff
@@ -0,0 +1,54 @@
...
+
+```yaml
+# model.yaml
+apiVersion: kubeai.org/v1
+kind: Model
+metadata:
+  name: my-model
+spec:
+  # ...
+  autoscalingProfile: "budget"
```
samos123, line 53: can you add an example that doesn't set autoScalingProfile and sets minReplicas directly?

internal/config/system.go
samos123: I think both is fine. 3 is good too as long as we're consistent throughout.

charts/kubeai/values.yaml
```diff
@@ -60,11 +60,26 @@ resourceProfiles:
   #   nodeSelector:
   #     cloud.google.com/gke-accelerator-type: tpu-v5-lite-podslice
 
+autoscaling:
+  interval: 10s
+  timeWindow: 10m
```
samos123, line 65: Can you add a comment about how interval and timeWindow work?

samos123: I often do `helm show values` to see what values are available and having doc strings helps a lot.

samos123 (approved):

docs/how-to/configure-autoscaling.md
nstogner: done

samos123:
Alternative:
```
apiVersion: kubeai.org/v1
kind: Model
metadata:
  name: faster-whisper-medium-en-cpu
spec:
  features: [SpeechToText]
  owner: Systran
  url: hf://Systran/faster-whisper-medium.en
  engine: FasterWhisper
  minReplicas: 0 # defaults to 0 if not set like before
  maxReplicas: 3 # defaults to 3 if not set like before
  concurrentRequests: 100 # defaults to 100 if not set
  ScaleDownDelay: 60s # defaults to 60s if not set
  resourceProfile: cpu:1
```

Benefits: simple and backwards compatible.

api/v1/model_types.go
```diff
@@ -63,6 +61,33 @@ type ModelSpec struct {
...
+	Autoscaling *ModelAutoscaling `json:"autoscaling,omitempty"`
+}
+
+type ModelAutoscaling struct {
+	// Disabled is used to disable autoscaling.
+	Disabled bool `json:"disabled,omitempty"`
+	// TargetRequests is the target number of active requests per Pod.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:default=100
+	TargetRequests int32 `json:"targetRequests"`
```
samos123, line 79: I think I prefer calling it concurrentRequest

charts/kubeai/values.yaml
nstogner: done

charts/kubeai/values-gke.yaml
```diff
@@ -0,0 +1,5 @@
+resourceProfiles:
+  nvidia-gpu-l4:
+    nodeSelector:
+      cloud.google.com/gke-accelerator: "nvidia-l4"
+      cloud.google.com/gke-spot: "true"
```
samos123, line 5: Remove spot node selector but add a toleration for spot

test/quickstart.sh
```diff
@@ -65,7 +65,7 @@ curl -s -X GET "http://localhost:8000/openai/v1/models" | jq '. | length == 0'
 # By using the --reuse-values flag we can just append models to the previous install
 # while avoiding overriding the image that skaffold originally built and set in the
 # first install.
-helm upgrade --reuse-values --install kubeai charts/kubeai -f - <<EOF
+helm upgrade --reuse-values --install kubeai ./charts/kubeai -f - <<EOF
```
samos123, line 68: Curious was this needed due to being on Mac?

samos123 (approved):

test/quickstart.sh
nstogner: No, this didnt solve anything - but it made it more clear for me that this was referencing a local dir

PR #216, opened 2024-09-17
Add support for OpenShift

amrut-asm:
OpenShift requires securityContext to be set for pods in order to decide an appropriate SCC to apply to them.

This PR:

1. Adds SecurityContext for the model pods created by kubeAI
2. Adds SecurityContext for the model containers created by kubeAI
3. Sets securityContext for openwebui container
4. Creates securityContextConstraints (SCC) for openwebui and the model pods

Chart version fields are untouched... not sure what to do with them

### DIFF
--- a/charts/kubeai/charts/openwebui/templates/securityContextConstraints.yaml
+++ b/charts/kubeai/charts/openwebui/templates/securityContextConstraints.yaml
@@ -1,0 +1,20 @@
+{{- if .Capabilities.APIVersions.Has "security.openshift.io/v1" }}
+{{- if .Values.serviceAccount.create }}
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: openwebui
+allowPrivilegeEscalation: false
+readOnlyRootFilesystem: false
+runAsUser:
+  type: RunAsAny
+seLinuxContext:
+  type: MustRunAs
+seccompProfiles:
+  - runtime/default
+requiredDropCapabilities:
+  - ALL
+users:
+  - system:serviceaccount:{{ .Release.Namespace }}:{{ include "openwebui.serviceAccountName" . }}
+{{- end }}
+{{- end }}
--- a/charts/kubeai/charts/openwebui/templates/securityContextConstraints.yaml
+++ b/charts/kubeai/charts/openwebui/templates/securityContextConstraints.yaml
@@ -1,9 +1,12 @@
+# Create securityContextConstraints for the openwebui pod if running on OpenShift.
+# This is needed because pods in OpenShift run with the restricted-v2 SCC by
+# default which do not allow a container to start with uid=0
+# (The openwebui image runs as the root user)
 {{- if .Capabilities.APIVersions.Has "security.openshift.io/v1" }}
-{{- if .Values.serviceAccount.create }}
 apiVersion: security.openshift.io/v1
 kind: SecurityContextConstraints
 metadata:
-  name: openwebui
+  name: {{ include "openwebui.fullname" . }}
 allowPrivilegeEscalation: false
 readOnlyRootFilesystem: false
 runAsUser:
@@ -16,5 +19,4 @@ requiredDropCapabilities:
   - ALL
 users:
   - system:serviceaccount:{{ .Release.Namespace }}:{{ include "openwebui.serviceAccountName" . }}
-{{- end }}
 {{- end }}
--- a/charts/kubeai/templates/_helpers.tpl
+++ b/charts/kubeai/templates/_helpers.tpl
@@ -62,6 +62,17 @@ Create the name of the service account to use
 {{- end }}
 
 {{/*
+Create the name of the service account to use for model pods
+*/}}
+{{- define "models.serviceAccountName" -}}
+{{- if .Values.modelServiceAccount.create }}
+{{- default (printf "%s-models" (include "kubeai.fullname" .)) .Values.modelServiceAccount.name }}
+{{- else }}
+{{- default "default" .Values.modelServiceAccount.name }}
+{{- end }}
+{{- end }}
+
+{{/*
 Create the name of the huggingface secret to use
 */}}
 {{- define "kubeai.huggingfaceSecretName" -}}
--- a/charts/kubeai/templates/configmap.yaml
+++ b/charts/kubeai/templates/configmap.yaml
@@ -22,9 +22,7 @@ data:
       securityContext:
         {{- .Values.modelServerPods.securityContext | toYaml | nindent 8}}
       {{- end}}
-      {{- if .Values.modelServiceAccount.create }}
       serviceAccountName: {{ include "models.serviceAccountName" . }}
-      {{- end}}
     {{- end }}
     modelAutoscaling:
       {{- .Values.modelAutoscaling | toYaml | nindent 6 }}
--- a/charts/kubeai/templates/configmap.yaml
+++ b/charts/kubeai/templates/configmap.yaml
@@ -12,6 +12,20 @@ data:
       {{- .Values.resourceProfiles | toYaml | nindent 6 }}
     modelServers:
       {{- .Values.modelServers | toYaml | nindent 6 }}
+    {{- if .Values.modelServerPods }}
+    modelServerPods:
+      {{- if .Values.modelServerPods.podSecurityContext }}
+      podSecurityContext:
+        {{- .Values.modelServerPods.podSecurityContext | toYaml | nindent 8}}
+      {{- end}}
+      {{- if .Values.modelServerPods.securityContext }}
+      securityContext:
+        {{- .Values.modelServerPods.securityContext | toYaml | nindent 8}}
+      {{- end}}
+      {{- if .Values.modelServiceAccount.create }}
+      serviceAccountName: {{ include "models.serviceAccountName" . }}
+      {{- end}}
+    {{- end }}
     modelAutoscaling:
       {{- .Values.modelAutoscaling | toYaml | nindent 6 }}
     messaging:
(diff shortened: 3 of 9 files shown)


nstogner:
Thanks for the PR! We will take a look.

api/v1/model_types.go
```diff
@@ -46,6 +47,15 @@ type ModelSpec struct {
 	// Will be set from ResourceProfile + Engine if not specified.
 	Image string `json:"image,omitempty"`
 
+	// Service Account to be used by the model pods
```
nstogner, line 50: We would like to keep the Model spec as simple as possible. I think it would be better to have these fields defined at the system-level (in the ConfigMap for KubeAI instead of the Model spec). Perhaps these would belong to a field inside of the `internal/config` pkg's `System` struct:

```go
type System struct {
    // ...
    ModelServerPods ModelServerPods `json:"modelServerPods"`
}
```

Let me know if there is any reason why defining the same values across all model Pods would be an issue.

nstogner (changes requested):
Thanks for taking the time to submit the PR! Added a comment about where we would prefer to specify these fields.

api/v1/model_types.go
samos123: What's your thought on this instead?
```golang
type System struct {
         // The service account to use for all model pods
	ModelServiceAccountName string `json:"serviceAccountName,omitempty"`

	// Security Context for the model pods
	ModelPodSecurityContext *corev1.PodSecurityContext `json:"podSecurityContext,omitempty"`

	// Security Context for the model pod containers
	ModelContainerSecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`
}
```

samos123:
Thank you for this PR! We think it would be very valuable to have OpenShift support and will try to merge this once changes are addressed.

The chart versions do not need to be bumped. We currently do manual releases with more testing. This allows us to merge changes into main without being afraid of breaking users.

It would also be very helpful to have an OpenShift specific installation guide once this PR gets merged or as part of this PR.

Feel free to also join the discord and we can discuss relatime if needed.

api/v1/model_types.go
samos123: I see and agree. I didn't  get your initial comment.

samos123:
Saw you closed the PR, but we're more than happy to take this PR for which already did most of the work and make the requested changes directly. You would still be author. Please let me know if you would prefer that. Apologies for not being able to take the PR as-is directly, because I do really think this would be valuable.

amrut-asm:
I discarded the commits and synced my branch with main and it somehow closed the PR 😔. I have made the changes already according to the new format, testing it currently. Will re-open this if possible otherwise I'll create a new PR 🙂

amrut-asm:
On the OpenShift specific installation guide - Nothing out of the ordinary... following the regular installation process will work fine

amrut-asm:
> Thank you for this PR! We think it would be very valuable to have OpenShift support and will try to merge this once changes are addressed.
> 
> The chart versions do not need to be bumped. We currently do manual releases with more testing. This allows us to merge changes into main without being afraid of breaking users.
> 
> It would also be very helpful to have an OpenShift specific installation guide once this PR gets merged or as part of this PR.
> 
> Feel free to also join the discord and we can discuss relatime if needed.


On the OpenShift specific installation guide - Nothing out of the ordinary... following the regular installation process will work fine

api/v1/model_types.go
amrut-asm: I've made the necessary modifications.

charts/kubeai/values.yaml
```diff
@@ -30,6 +30,17 @@ modelServers:
       default: "fedirz/faster-whisper-server:latest-cpu"
       nvidia-gpu: "fedirz/faster-whisper-server:latest-cuda"
 
+modelServerPods:
+  # Security Context for the model pods
+  # Needed for OpenShift
+  securityContext:
```
samos123, line 36: Just curious, no change needed. Have you seen this cause issues for pods?

I don't expect the LLM engines to fail but there is always a risk of issues when you enable more security settings by default. 

I do like having secure defaults that work! So I'm in favor of keeping this as-is.

charts/kubeai/templates/configmap.yaml
```diff
@@ -12,6 +12,20 @@ data:
...
+    modelServerPods:
+      {{- if .Values.modelServerPods.podSecurityContext }}
+      podSecurityContext:
+        {{- .Values.modelServerPods.podSecurityContext | toYaml | nindent 8}}
+      {{- end}}
+      {{- if .Values.modelServerPods.securityContext }}
+      securityContext:
+        {{- .Values.modelServerPods.securityContext | toYaml | nindent 8}}
+      {{- end}}
+      {{- if .Values.modelServiceAccount.create }}
```
samos123, line 25: What should be the behavior if someone sets `.Values.modelServiceAccount.name` and creates this K8s SA themselves?

I think in that case the serviceAccountName should be set.

I think this if should be removed and then we need to ensure that `.Values.modelServiceAccount.name` is `default` and `create=false` by default.

samos123: Note we do this as a follow up because create=true by default. So this is more of an edge case.

charts/kubeai/templates/securityContextConstraints.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Capabilities.APIVersions.Has "security.openshift.io/v1" }}
+{{- if .Values.modelServiceAccount.create -}}
```
samos123, line 2: same comment in regards to setting custom service account and create=false

samos123 (approved):
I left some comments but LGTM overall to merge and we can tackle these as follow ups or you could.

Thank you for adding OpenShift support and improving out of the box security! Will wait for @nstogner to finish his review too.

charts/kubeai/charts/openwebui/templates/securityContextConstraints.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Capabilities.APIVersions.Has "security.openshift.io/v1" }}
+{{- if .Values.serviceAccount.create }}
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: openwebui
```
nstogner, line 6: Let's template this with fullname like other resources.

charts/kubeai/templates/securityContextConstraints.yaml
```diff
@@ -0,0 +1,20 @@
...
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: kubeai-models
```
nstogner, line 6: Let's template this with fullname like other resources.

nstogner (changes requested):
Looking good - a minor things.

amrut-asm:
> Looking good - a minor things.

I'll address the issues and fix them soon 🙂

charts/kubeai/templates/configmap.yaml
amrut-asm: > What should the behavior be if someone sets `.Values.modelServiceAccount.name` with `create=false` and creates this K8s SA themselves without using helm?
> 
> I think in that case the serviceAccountName should be set.
> 
> I think this if statement could be removed?

Yes, I agree. This statement is not needed.

charts/kubeai/templates/securityContextConstraints.yaml
amrut-asm: Agree

charts/kubeai/templates/_helpers.tpl
```diff
@@ -61,6 +61,17 @@ Create the name of the service account to use
...
 {{- end }}
 
+{{/*
+Create the name of the service account to use for model pods
+*/}}
+{{- define "models.serviceAccountName" -}}
+{{- if .Values.modelServiceAccount.create }}
+{{- default (printf "%s-models" (include "kubeai.fullname" .)) .Values.modelServiceAccount.name }}
+{{- else }}
+{{- default "default" .Values.modelServiceAccount.name }}
```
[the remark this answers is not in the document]
amrut-asm: Sorry, I don't understand. It is defaulting to "default" when 
1. modelServiceAccount.create is false
2. modelServiceACcount.name is empty ""

Is it not expected?

charts/kubeai/templates/securityContextConstraints.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Capabilities.APIVersions.Has "security.openshift.io/v1" }}
```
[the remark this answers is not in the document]
amrut-asm: Done

charts/kubeai/values.yaml
amrut-asm: > Just curious, no change needed. Have you seen this cause issues for pods in general (not on KubeAI)?
> 
> I don't expect the LLM engines to fail but there is always a risk of issues when you enable more security settings by default.
> 
> I do like having secure defaults that work! So I'm in favor of keeping this as-is.

On OpenShift, from what I've observed - a container must set it's securityContext according to the privileges it requires so as to be assigned an appropriate SCC that matches it's securityContext settings. Failing to do so (for example, not setting securityContext at all or not matching the securityContext for the SCC which we are *hoping* to be assigned) will result in the default SCC to be assigned (which happens to be restricted-v2).

So, yeah this causing issues for pods is only true on OpenShift and not for others.

samos123: Thank a lot for explaining. I get it now.

charts/kubeai/charts/openwebui/templates/securityContextConstraints.yaml
amrut-asm: Done

charts/kubeai/templates/securityContextConstraints.yaml
amrut-asm: Done

charts/kubeai/templates/_helpers.tpl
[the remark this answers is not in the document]
samos123: It's correct as-is: https://helm.sh/docs/chart_template_guide/function_list/#default

no need to change it.

samos123 (approved):

PR #220, opened 2024-09-18
Separate charts [BREAKING CHANGE TO HELM CHARTS]

### LINKED ISSUE #213: Helm upgrade does not upgrade CRDs
Current method of specifying CRDs is ignored when upgrading helm release.

See: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

Possible solution:

* Move CRDs to `./charts/kubeai/templates`
* Move model up a level to be its own chart

Bonus:

* Template a catalog of individual `<model-name>.yaml` files that could be installed with `kubectl apply -f https://...` directly. Probably makes sense to do this with the models chart.

nstogner:
Split the model catalog in `charts/kubeai/` into its own chart (`charts/models/`). Move CRDs to `templates/` to allow users to `helm upgrade` CRDs.

Fixes #213


docs/README.md
```diff
@@ -58,23 +58,28 @@ helm repo update
...
-EOF
-
 helm upgrade --install kubeai kubeai/kubeai \
-    -f ./helm-values.yaml \
     --wait --timeout 10m
 ```
 
+Install some predefined models.
+
+```bash
```
samos123, line 67: another alternative could be to use `kubectl apply -f http://`

I kind of like the idea that you always see the full model CRD that is getting applied.

nstogner: I like that idea too. Any customizations would likely need to be done with a following kubectl command or a multi-step download file, edit, apply... Thinking of setting `minReplicas: 1`

nstogner: Lets punt on this for now.

charts/kubeai/Chart.yaml
```diff
@@ -41,7 +29,7 @@ home: https://www.substratus.ai
 
 maintainers:
 - name: nstogner
-  email: nicholas.stogner@gmail.com
+  url: https://www.linkedin.com/in/nstogner/
```
samos123, line 32: I think we need an email later if we upload to artifact hub, but fine with removing for now

docs/installation/gke.md
```diff
@@ -39,33 +39,44 @@ TODO: Reference gcloud commands for creating a GKE standard cluster.
...
+## 3. Optionally configure models
 
-Install KubeAI with [Helm](https://helm.sh/docs/intro/install/).
+Optionally install preconfigured models.
 
 ```bash
-helm repo add kubeai https://substratusai.github.io/kubeai/
-helm repo update
+cat <<EOF > kubeai-models.yaml
+models:
```
samos123, line 74: should this now be:
```
catalog:
  llama-..:
```

docs/installation/gke.md
```diff
@@ -39,33 +39,44 @@ TODO: Reference gcloud commands for creating a GKE standard cluster.
...
-Define the installation values for GKE.
+Add KubeAI Helm repository.
 
 ```bash
-cat <<EOF > helm-values.yaml
-models:
-  catalog:
-    llama-3.1-8b-instruct-fp8-l4:
-      enabled: true
+helm repo add kubeai https://substratusai.github.io/kubeai/
```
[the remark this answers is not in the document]
samos123: I am fixing this in a follow up commit

docs/installation/gke.md
samos123: I am fixing this in a follow up commit

samos123 (approved):

PR #226, opened 2024-09-19
fix helm release process and bump models chart version

samos123:
Currently the models helm chart isn't being released automatically this is a try at fixing that.


samos123:
This fixes it, but at some point we will probably have to improve this. I am not sure if the workaround is really needed.

PR #230, opened 2024-09-20
add llama 3.1 70b and h100 resourceProfile

samos123:


charts/kubeai/values.yaml
```diff
@@ -49,6 +49,16 @@ resourceProfiles:
         operator: "Equal"
         value: "present"
         effect: "NoSchedule"
+  nvidia-gpu-h100:
```
nstogner, line 52: Can you also add the GKE-specific node selectors to the GKE install guide / the GKE values file?

Other than that - LGTM - feel free to merge

nstogner (changes requested):

charts/kubeai/values.yaml
samos123: done

PR #233, opened 2024-09-21
fix #232 secret token empty validation error

### LINKED ISSUE #232: Error when installing kubeai
Install kubeai according to the steps prompted by the official website documentation. When proceeding to helm installation, the error message cannot be installed as shown in the figure below. How to solve it?
error

samos123:
Some users have reported validation issues with a secret that has an empty string as value.

This isn't reproducible on Kind and latest versions. So this fix hasn't been validated.

### DIFF
--- a/charts/kubeai/templates/secret.yaml
+++ b/charts/kubeai/templates/secret.yaml
@@ -1,4 +1,6 @@
-{{- if .Values.secrets.huggingface.create -}}
+# Only create the secret if the token is not empty.
+# See: https://github.com/substratusai/kubeai/issues/232
+{{- if and .Values.secrets.huggingface.create (not (empty .Values.secrets.huggingface.token)) -}}
 apiVersion: v1
 kind: Secret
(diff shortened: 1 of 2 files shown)


charts/kubeai/templates/secret.yaml
```diff
@@ -1,4 +1,6 @@
-{{- if .Values.secrets.huggingface.create -}}
+# Only create the secret if the token is not empty.
+# See: https://github.com/substratusai/kubeai/issues/232
+{{- if and .Values.secrets.huggingface.create (not (empty .Values.secrets.huggingface.token)) -}}
```
nstogner, line 3: This conditional should be for the .data.value IMO, not the whole secret

samos123: You mean remove the and?

samos123: It's already for the data right now

Issue #231, opened 2024-09-20
Issue: autoscale print targetrequests value instead of the pointer address


benz9527:
![image](https://github.com/user-attachments/assets/329e4935-a94f-4c63-8d97-e822de728a6e)

![image](https://github.com/user-attachments/assets/641292b8-a38b-438c-93c2-772174237c91)

nstogner:
Good catch! If you would like to be added as a contributor to the project feel free to submit a PR to tackle this (`m.Spec.TargetRequests` should never be nil in this case so it should just be fixed by de-referencing: `*m.Spec.TargetRequests`).

Otherwise, we will fix soon.

benz9527:
I glad to make a PR.

benz9527:
I try to push a new fix branch which is created from main, but I found that I haven't permission to create a remote branch. Haha.

samos123:
You can create a fork and create a branch in your fork. Afterwards you can submit a PR.

benz9527:
> You can create a fork and create a branch in your fork. Afterwards you can submit a PR.

Oh, I forget that github opensource project need to  be forked and then I can PR.

PR #239, opened 2024-09-23
fix: #231 Print target requests value

benz9527:
The `m.Spec.TargetRequests` in this case will never be nil, because it can be set by default value as 100. So I just fix it by de-referencing.


samos123:
Thanks @benz9527 for taking this in on yourself! Greatly appreciated.

PR #240, opened 2024-09-23
fix #235 utilize standard k8s labels

### LINKED ISSUE #235: Utilize k8s standard labels on model pods and add more labels
See here for recommended labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

I suggest using the following:
```
app.kubernetes.io/name: vllm
app.kubernetes.io/instance: vllm-llama-3.1-7b
app.kubernetes.io/managed-by: kubeai
app.kubernetes.io/component: engine
```

This is getting important since we're looking at prometheus pod scraping configs that target the labels. So I want to standardize on the labels before implementing that.

samos123:


internal/modelcontroller/model_controller.go
```diff
@@ -737,7 +737,14 @@ func (r *ModelReconciler) infinityPodForModel(m *kubeaiv1.Model, profile ModelCo
 }
 
 func labelsForModel(m *kubeaiv1.Model) map[string]string {
-	return map[string]string{"app": "model", "model": m.Name}
+	return map[string]string{
+		"app":                          "model",
+		"model":                        m.Name,
+		"app.kubernetes.io/name":       "vllm",
```
samos123, line 743: This is incorrect. I need to get engine from model.

nstogner (approved):
LGTM

PR #248, opened 2024-09-26
rough min model replicas

sam-huang1223:


nstogner:
Thanks for the PR Sam! I discussed this with other-Sam offline and we came to the conclusion it makes the most sense to put this into the `models` helm chart:

```yaml
all:
  podDisruptionBudget: # Would apply to all models via a label selector that matches any model that has a "model" key
    create: false
    # ... values for availability
```

PR #249, opened 2024-09-26
Initial TPU support

nstogner:
* Add v5e resource profiles for GKE
* Add models: `gemma-2b-it-tpu`, `llama-3.1-8b-instruct-tpu`
* Remove outdated vLLM metrics workaround (conflicted with using flags needed to support multiple TPUs)
* Update vLLM cpu image to include metrics fix from vLLM

### DIFF
--- a/charts/kubeai/values.yaml
+++ b/charts/kubeai/values.yaml
@@ -18,7 +18,7 @@ modelServers:
       # The "default" image should always be specified.
       # "default" is used when no imageName is specified or if a specific image is not found.
       default: "vllm/vllm-openai:v0.6.1.post2"
-      cpu: "substratusai/vllm:v0.5.5-cpu"
+      cpu: "ghcr.io/substratusai/vllm:v0.6.1-cpu"
       nvidia-gpu: "vllm/vllm-openai:v0.6.1.post2"
       google-tpu: "ghcr.io/substratusai/vllm:v0.6.1-tpu"
   OLlama:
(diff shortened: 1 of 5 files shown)


charts/kubeai/values.yaml
```diff
@@ -18,10 +18,9 @@ modelServers:
       # The "default" image should always be specified.
       # "default" is used when no imageName is specified or if a specific image is not found.
       default: "vllm/vllm-openai:v0.6.1.post2"
-      cpu: "substratusai/vllm:v0.5.5-cpu"
+      cpu: "ghcr.io/substratusai/vllm:v0.6.1-cpu"
```
samos123, line 21: hmm these images tend to be much slower for me. I think in the past it was like 10x slower than using Docker Hub on GKE. GCP seems to cache Docker hub automatically but not ghcr.io

charts/kubeai/values.yaml
```diff
@@ -18,10 +18,9 @@ modelServers:
...
       nvidia-gpu: "vllm/vllm-openai:v0.6.1.post2"
-      # TODO publish tpu image.
-      google-tpu: "substratusai/vllm-openai-tpu:v0.5.5"
+      google-tpu: "ghcr.io/substratusai/vllm:v0.6.1-tpu"
```
samos123, line 23: same

charts/models/values.yaml
```diff
@@ -21,6 +22,36 @@ catalog:
...
+    url: "hf://google/gemma-2b-it"
+    engine: VLLM
+    resourceProfile: google-tpu-v5e-1x1:1
+    args:
+    - --disable-log-requests
+  # gemma2-9b-it-fp8-tpu:
+  #   enabled: false
+  #   features: ["TextGeneration"]
+  #   owner: neuralmagic
+  #   # vLLM logs: "ValueError: fp8 quantization is currently not supported in TPU Backend."
```
samos123, line 38: you need to use the 16 bit version probably

samos123 (approved):

PR #254, opened 2024-09-27
add llama 3.1 405b model

samos123:


charts/models/values.yaml
```diff
@@ -117,12 +117,30 @@ catalog:
...
       - --max-model-len=65536
       - --max-num-batched-token=65536
+      - --max-num-seqs=1024
       - --gpu-memory-utilization=0.9
       - --tensor-parallel-size=2
       - --enable-prefix-caching
       - --disable-log-requests
     resourceProfile: nvidia-gpu-h100:2
     targetRequests: 500
+  llama-3.1-405b-instruct-fp8:
```
nstogner, line 127: Other models use the convention of suffixing with the accelerator: see above: `llama-3.1-70b-instruct-fp8-h100:`

nstogner (changes requested):
One naming convention change - otherwise lgtm

charts/models/values.yaml
samos123: so I purposely did this with the anticipation of adding a100-80b resource profile once we support multi resourceProfile

samos123: I rather start using this convention now for any new models but keep existing models as-is since users may be using them already.

samos123: Done

PR #259, opened 2024-09-29
add integration test for multiple slash

samos123:


test/quickstart.sh
```diff
@@ -87,9 +87,21 @@ wait_for_pod_ready model=gemma2-2b-cpu
...
+  if [ "$http_code" -ne 200 ]; then
+    echo "Failed to get completions from $url"
+    echo "HTTP code: $http_code"
+    exit 1
+  fi
+}
+
+test_completion "http://localhost:8000/openai/v1/completions"
+# Double slash in the URL should work as well.
+test_completion "http://localhost:8000/openai//v1/completions"
```
alpe, line 104: This fails with 'HTTP code: 301'

You can use a curl param to follow the redirect or just assert the 301 with the correct target
```
-L, --location
              (HTTP) If the server reports that the requested page has moved to a different location (indicated with a Location: header and a 3XX response code), this option
              makes curl redo the request on the new place. If used together with -i, --include or -I, --head, headers from all requested pages are shown.
```

samos123: Thanks for catching this! This now correctly follows and shows the 400 error.

samos123:
@alpe if you have some ideas on how to fix this, feel free to just take my test and submit a PR. I haven't started looking into how to fix it yet.

test/quickstart.sh
samos123: weird, on my local curl I can reproduce 400 error, but seems the automated test does not.

alpe:
👍 The test passes with the `-L` flag, now.

PR #263, opened 2024-10-06
Refactor e2e tests

### LINKED ISSUE #262: Remove request body logging
https://github.com/substratusai/kubeai/blob/884aa62c81e2f5575a92c60a62f6e8734418a17b/internal/modelproxy/request.go#L131

This becomes especially problematic with binary requests (Speech to Text).

nstogner:
* Add central bash script to run e2e test cases
* Separate e2e tests out into separate cases
* Add python openai client test case
* Fix #262


skaffold.yaml
```diff
@@ -5,20 +5,39 @@ metadata:
...
     push: false
 deploy:
   helm:
     releases:
     - name: kubeai
-      chartPath: charts/kubeai
-      valuesFiles:
-      - charts/kubeai/values.yaml
-      setValueTemplates:
-        image.tag: "{{.DIGEST_HEX}}"
```
samos123, line 18: i think i needed this to ensure it uses the current revision

nstogner: `skaffold render` shows that it mutates the image tag itself: `image: substratusai/kubeai:helm-chart-kubeai-0.6.0-4-gcdeaea1`

nstogner:
End-to-end tests run in parallel (in sequence after unit and integration tests - if those fail, theres not use in tying up resources to run e2e tests):
Screenshot 2024-10-07 at 8 25 35 PM

nstogner:
Should be good to merge if all looks OK

samos123:
Nice! LGTM.

I think I rather run e2e and unit /integration test at the same time so we don't have to wait an additional 4 minutes. I can tackle that in a follow up, lets merge this first.

.github/workflows/tests.yml
```diff
@@ -8,38 +8,42 @@ on:
...
 
-      - name: Run unit & integration tests
-        run: make test
-  e2e-quickstart:
+      - name: Run integration tests
+        run: make test-integration
+
+  e2e:
     runs-on: ubuntu-latest
+    needs: unit-and-integration # No use in running e2e tests if integration tests fail.
```
samos123, line 25: I would prefer to remove this needs

samos123: if we start running out of resources, then we can consider adding it back.

samos123 (approved):

.github/workflows/tests.yml
nstogner: done

PR #268, opened 2024-10-09
Add Autoscaler State ConfigMap

### LINKED ISSUE #265: Autoscaler should maintain state across KubeAI restarts & crashes
KubeAI should set the average requests per model to the last remembered for the restart.

nstogner:
Fixes #265

### DIFF
--- a/charts/kubeai/values.yaml
+++ b/charts/kubeai/values.yaml
@@ -109,7 +109,7 @@ modelAutoscaling:
   timeWindow: 10m
   # The name of the ConfigMap that stores the state of the autoscaler.
   # Defaults to "{fullname}-autoscaler-state".
-  # stateConfigMapName: "kubeai-autoscaler-state"
+  stateConfigMapName: ""
 
 messaging:
   errorMaxBackoff: 30s
--- a/charts/kubeai/values.yaml
+++ b/charts/kubeai/values.yaml
@@ -107,6 +107,9 @@ modelAutoscaling:
   # Time window the autoscaling algorithm will consider when calculating
   # the desired number of replicas.
   timeWindow: 10m
+  # The name of the ConfigMap that stores the state of the autoscaler.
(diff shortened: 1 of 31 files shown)


charts/kubeai/values.yaml
```diff
@@ -107,6 +107,9 @@ modelAutoscaling:
   # Time window the autoscaling algorithm will consider when calculating
   # the desired number of replicas.
   timeWindow: 10m
+  # The name of the ConfigMap that stores the state of the autoscaler.
+  # Defaults to "{fullname}-autoscaler-state".
+  # stateConfigMapName: "kubeai-autoscaler-state"
```
samos123, line 112:
```suggestion
  stateConfigMapName: ""
```

samos123: this is what we do for other names like secrets.huggingface.name

internal/config/system.go
```diff
@@ -62,6 +62,9 @@ func (s *System) DefaultAndValidate() error {
 	if s.ModelAutoscaling.TimeWindow.Duration == 0 {
 		s.ModelAutoscaling.TimeWindow.Duration = 10 * time.Minute
 	}
+	if s.ModelAutoscaling.StateConfigMapName == "" {
+		s.ModelAutoscaling.StateConfigMapName = "kubeai-autoscaler-state"
```
samos123, line 66: I think we should remove that check and fail if it's not set, since it should always be set by the default value from helm.

charts/kubeai/values.yaml
nstogner: Added

internal/config/system.go
nstogner: Done

samos123 (approved):

nstogner:
Found an issue - the state gets forgotten if there are no active requests for the model

PR #271, opened 2024-10-11
add tpu quota to GKE install guide and use values-gke.yaml

samos123:


docs/installation/gke.md
```diff
@@ -51,16 +54,9 @@ helm repo update
...
 ```bash
-cat <<EOF > kubeai.yaml
-resourceProfiles:
-  nvidia-gpu-l4:
-    nodeSelector:
-      cloud.google.com/gke-accelerator: "nvidia-l4"
-      cloud.google.com/gke-spot: "true"
-EOF
-
+curl -L -O https://raw.githubusercontent.com/substratusai/kubeai/refs/heads/main/charts/kubeai/values-gke.yaml
```
nstogner, line 57: Prefer to avoid the `curl` by updating the helm flag `-f https://raw.githubusercontent.com/substratusai/kubeai/refs/heads/main/charts/kubeai/values-gke.yaml`

docs/installation/gke.md
```diff
@@ -8,12 +8,15 @@ There are 3 critical quotas you will need to verify for this guide. The minimum
 
 | Quota                      | Location      | Min Value |
 |----------------------------|---------------|-----------|
+| Preemptible TPU v5 Lite Podslice chips | `<your-region>` | 8 |
```
nstogner, line 11: We should call out somewhere that you really only need TPU or GPU quota depending on what you are trying to do.

nstogner (changes requested):

docs/installation/gke.md
samos123: I started with that, but the url is too long and I felt like it hides away too much then. I prefer the curl so people can easily edit later and have it saved somewhere.

nstogner: Sounds good

docs/installation/gke.md
samos123: good call, I updated the description above the table

nstogner (approved):
Looks good!

PR #281, opened 2024-10-20
Adding Build WF timeout to address stuck WF's

Sudhamsh:
This change will limit the WF action timeout to address GitHub hosted action communication issue.

### DIFF
--- a/.github/workflows/build-push-kubeai.yml
+++ b/.github/workflows/build-push-kubeai.yml
@@ -17,8 +17,8 @@ env:
 jobs:
   kubeai:
     runs-on: ubuntu-latest
-    # Timeout to address Github actions stuck scenarios 
-    timeout-minutes: 10
+    # Timeout to address Github actions stuck scenarios.
+    timeout-minutes: 60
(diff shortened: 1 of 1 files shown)


samos123:
Note, this isn't the github action that's stuck. The github action that's stuck is managed by Github themselves since it's for the Github pages publishing workflow. So sadly this wouldn't help for that case. The default timeout (6 hours) is enough for our kubeai managed github actions.

.github/workflows/build-push-kubeai.yml
```diff
@@ -17,6 +17,8 @@ env:
 jobs:
   kubeai:
     runs-on: ubuntu-latest
+    # Timeout to address Github actions stuck scenarios 
+    timeout-minutes: 10
```
samos123, line 21: Let's increase this to 60 minutes because the action can take up to 20 minutes in normal case.

samos123: I added a commit that does this.

samos123 (approved):

samos123:
Thanks for taking the time to make the PR and helping with addressing stuck workflows!

PR #284, opened 2024-10-24
add kubeai metrics service endpoint

kaiehrhardt:
to be able to fetch it via serviceMonitor

### DIFF
--- a/charts/kubeai/Chart.yaml
+++ b/charts/kubeai/Chart.yaml
@@ -7,7 +7,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.7.0
+version: 0.7.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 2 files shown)


charts/kubeai/templates/service.yaml
```diff
@@ -11,5 +11,9 @@ spec:
       targetPort: http
       protocol: TCP
       name: http
+    - port: 8080
```
samos123, line 14:
```suggestion
    - port: {{ .Values.service.metricsPort }}
```

and in values.yaml set default to "8080"

samos123: The metrics port is configurable in KubeAI so we should also expose this in the service object.

charts/kubeai/Chart.yaml
```diff
@@ -7,7 +7,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.7.0
+version: 0.7.1
```
samos123, line 10: note, you can leave this in this PR, but in the future, no need to bump the chart versions. I handle this as part of releases.

charts/kubeai/templates/service.yaml
samos123: actually ignore this, we don't have a way to override this anyway in the helm chart. Resolving this comment.

PR #294, opened 2024-10-27
Deep Chat integration

### LINKED ISSUE #293: Add guide for using DeepChat
See [DeepChat](https://github.com/OvidijusParsiunas/deep-chat/issues/288) issue.

nstogner:
See: https://github.com/OvidijusParsiunas/deep-chat

Fixes #293


examples/private-deep-chat/static/index.html
```diff
@@ -0,0 +1,19 @@
...
+<head>
+    <meta charset="UTF-8" />
+</head>
+<script type="module" src="https://unpkg.com/deep-chat@2.0.1/dist/deepChat.bundle.js"></script>
+
+<body>
+    <deep-chat id="chat-element" style="border-radius: 10px; width: 96vw; height: calc(100vh - 70px); padding-top: 10px"
+        messageStyles='{"default": {"shared": {"innerContainer": {"fontSize": "1rem"}}}}'
+        inputAreaStyle='{"fontSize": "1rem"}'
+        connect='{"url":"/openai/v1/chat/completions", "credentials": "same-origin", "stream": true}'
```
nstogner, line 13: Looks like setting `"credentials": "same-origin"` or `"include"` does not influence Deep Chat to send the basic auth header.

See Deep Chat docs:
https://deepchat.dev/docs/connect#connect-1

[Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials) linked to from Deep Chat docs seem to indicate it should:
```
Credentials are cookies, [TLS](https://developer.mozilla.org/en-US/docs/Glossary/TLS) client certificates, or authentication headers containing a username and password.
```

nstogner: It appears that the OpenAI key is sent instead:

Screenshot 2024-10-29 at 11 35 35 AM

nstogner: Solved via using a request interceptor, see https://github.com/OvidijusParsiunas/deep-chat/issues/288#issuecomment-2446402249

PR #300, opened 2024-11-01
Add gh200 support and model

happytreees:
This PR will add the initial support for GH200 nodes in KubeAI

### DIFF
--- a/charts/kubeai/values.yaml
+++ b/charts/kubeai/values.yaml
@@ -109,8 +109,6 @@ resourceProfiles:
         operator: "Equal"
         value: "present"
         effect: "NoSchedule"
-    nodeSelector:
-      nvidia.com/gpu.family: "hopper"
   nvidia-gpu-a100-80gb:
     imageName: "nvidia-gpu"
     limits:
--- a/charts/kubeai/values.yaml
+++ b/charts/kubeai/values.yaml
@@ -20,6 +20,7 @@ modelServers:
       default: "vllm/vllm-openai:v0.6.3.post1"
       cpu: "substratusai/vllm:v0.6.3.post1-cpu"
       google-tpu: "substratusai/vllm:v0.6.3.post1-tpu"
+      gh200: "drikster80/vllm-gh200-openai:v0.6.3.post1"
   OLlama:
     images:
       default: "ollama/ollama:latest"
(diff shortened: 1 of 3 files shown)


charts/kubeai/values.yaml
```diff
@@ -20,6 +20,7 @@ modelServers:
       default: "vllm/vllm-openai:v0.6.3.post1"
       cpu: "substratusai/vllm:v0.6.3.post1-cpu"
       google-tpu: "substratusai/vllm:v0.6.3.post1-tpu"
+      gh200: "drikster80/vllm-gh200-openai:v0.6.3.post1"
```
samos123, line 23: the plan would be to remove this once the default image works with `gh200`. I do like how you created an image just for gh200 for now.

charts/kubeai/values.yaml
```diff
@@ -97,6 +98,19 @@ resourceProfiles:
...
+    limits:
+      nvidia.com/gpu: "1"
+    requests:
+      nvidia.com/gpu: "1"
+    tolerations:
+      - key: "nvidia.com/gpu"
+        operator: "Equal"
+        value: "present"
+        effect: "NoSchedule"
+    nodeSelector:
```
samos123, line 112: could you remove this nodeSelector for the default values.yaml?  The reason is because this wouldn't work for users that aren't using the same node labels. We try to move any environment specific nodeSelectors to separate values files, see for example values-gke.yaml and values-eks.yaml. 

Let's create a values-nvidia-device-plugin.yaml that adds these kind of nodeSelectors.  It seems anyone using the values-nvidia-device-plugin could technically use these. Note that for GKE NAP and EKS karpetner this specific nodeSelector wouldn't work as expected.

charts/models/values.yaml
```diff
@@ -161,6 +161,18 @@ catalog:
...
+    features: [TextGeneration]
+    url: hf://hugging-quants/Meta-Llama-3.1-70B-Instruct-AWQ-INT4
+    engine: VLLM
+    args:
+      - --max-model-len=16384
+      - --max-num-batched-token=16384
+      - --enable-prefix-caching
+      - --disable-log-requests
+    resourceProfile: nvidia-gpu-gh200:1
+    targetRequests: 50
```
samos123, line 175: Let's mergethis as-is. We can optimize later. 

Here are the are the settings I would suggest trying out as well:
```
  args:
      - --max-num-seqs=1024
      - --kv-cache-dtype=fp8
      - --enforce-eager
      - --enable-chunked-prefill
      - --num-scheduler-steps=8
targetRequests: 1024
```
not sure if those settings work with awq, but I've had great luck with the fp8 models published by Neural Magic. You should be able to fit fp8 model on a single gh200 and have decent context length if you enable --kv-cache-dtype=fp8

samos123 (changes requested):
Thanks for the PR! Exciting to see GH200 being used.

Let's get this merged after you remove the nodeSelector from values.yaml. I left a few comments.

charts/kubeai/values.yaml
happytreees: Removed the nodeselector

charts/models/values.yaml
happytreees: I attempted to use an FP8 with the provided args, but it is not working properly. I think it might be best to add that model separately outside of this PR.

samos123 (approved):

Issue #303, opened 2024-11-03
how to access model files by pvc?


qiankunli:
Assuming all model files are stored on OSS/bucket, PVs are provided through StorageClass（related with oss）, and each model pod is associated with a PVC. Within the pod, VLLM can access model files by accessing a mounted directory( like`/data/models`).

How should the above mode be configured?

nstogner:
Hey there, thanks for submitting an issue!

Currently KubeAI supports loading models from non volume locations (i.e. huggingface hub). We do support caching models to volumes (See [caching docs](https://www.kubeai.org/concepts/storage-caching/#b-model-on-shared-filesystem-read-write-many)). With the current caching implementation, KubeAI manages the PVC itself and always downloads the model from elsewhere (i.e. huggingface hub).

It sounds like you want to define a PVC / PV as the source of the model (not just a caching location)? If so, can you elaborate a little more on your system design?

qiankunli:
The following diagram is my understanding of KubeAI (there may be errors).

![whiteboard_exported_image (4)](https://github.com/user-attachments/assets/9e8d8b33-e5a9-4e29-87d0-7ae577fecff4)

China cannot directly access Hugging Face, so we usually download the model file with proxy and then upload it to a bucket on OSS. There is a corresponding StorageClass for OSS to provide PV and PVC for the bucket, allowing the pod to access the model files through the PVC, as if the model files were in a directory within the pod.

![whiteboard_exported_image (4)](https://github.com/user-attachments/assets/0b84cc46-aae2-443e-9ea4-2c2172a9dab2)

samos123:
@qiankunli curious if this would help with your use case as well? https://github.com/substratusai/kubeai/issues/311

nstogner:
Adding direct support for `oss://` urls here: #304

nstogner:
OSS is now supported: https://www.kubeai.org/how-to/authenticate-to-model-repos/#alibaba-object-storage-service

samos123:
@qiankunli  Storing models on a PVC is now supported with vLLM engine. Please update your helm chart to v0.10.0 or later to try it out.

PR #304, opened 2024-11-04
LoRA Adapters for vLLM & support for s3, gs, oss for pulling adapters and models (to cache) from buckets

### LINKED ISSUE #132: Support dynamic LoRA serving

nstogner:
* Add `.spec.adapters`
* Support requesting adapters using the pattern: `{"model": "<model>_<adapter>", ... }`
* Load LoRA adapters into running vLLM containers
* Support updating LoRA adapters without needing to restart vLLM
* Rewrite `.model` to use adapter in chat request body when proxying to vLLM
* Add adapters to model list
* Add support for `s3://`, `gs://`, `oss://` urls (for adapters and cache loading)
* Add new cloud credentials to support new urls
* Update docs
* Update Model validation

NOTE:
* Was unable to test `oss://` urls... Had issues opening acct.

FOLLOWUP:
* Need to add adapter e2e tests (have not found a small enough model with adapters for use in kind cluster)
* Need to update chart values.yaml to include GH-actions-built image for model loader after merge!!!

Fixes #132, #303


alpe:
Nice drawing. This is very helpful! 🌻 
I am not super familiar with LoRa adapters but they can have significant size from what I saw. Caching seems a good idea. For the non-cache scenario, I would suggest to have `no-cache` or  `container-managed` profile so that it does not look like the default to skip this. 
With on-demand LoRa, the disk size may become a problem at some point. This is off scope but purge job or retention time can be things that need to be configured at some point in the profile.

samos123:
Can you show an example that has the url field? I'm assuming the url field must be used to specify the base model?

nstogner:
@samos123 I currently have all examples in the diagrams

nstogner:
Model `.spec.url` would be the same as normal.

nstogner:
Note, it looks like vLLM supports loading adapters from huggingface: https://github.com/vllm-project/vllm/pull/6234

charts/kubeai/templates/configmap.yaml
```diff
@@ -14,7 +14,7 @@ data:
       {{- .Values.cacheProfiles | toYaml | nindent 6 }}
     modelServers:
       {{- .Values.modelServers | toYaml | nindent 6 }}
-    modelLoaders:
+    modelLoading:
```
samos123, line 17: `modelLoaders` sounds better to me

hack/dev-models/gke-vllm-gpu-adapters.yaml
```diff
@@ -0,0 +1,39 @@
...
+    model-pod-ip: "127.0.0.1"
+    model-pod-port: "7000" 
+spec:
+  features: [TextGeneration]
+  owner: meta-llama
+  url: hf://TinyLlama/TinyLlama-1.1B-Chat-v0.3
+  #url: hf://meta-llama/Llama-2-7b
+  adapters:
+  - id: test
+    url: hf://jashing/tinyllama-colorist-lora
```
samos123, line 17: does vLLM support directly loading this adapter from HF or is it a hard requirement to download the lora adapter first?

nstogner: vLLM can load it from HF but not S3

nstogner: See: https://github.com/vllm-project/vllm/blob/d1c6799b8870e513bf4f2305cbf6cda9fc3d773b/vllm/lora/utils.py#L178

samos123:
Not sure whether this should work or not. So sharing what I did and the error message I get:
```
git checkout $THIS_PR

k get pods
NAME                         READY   STATUS             RESTARTS     AGE
kubeai-6bf98d5b77-txmpb      0/1     CrashLoopBackOff   1 (2s ago)   5s
openwebui-69ffb7dbb4-xcvlj   0/1     Running            0            5s

k logs -f kubeai-6bf98d5b77-txmpb 
2024-11-24T06:37:29Z    INFO    manager run finished
2024-11-24T06:37:29Z    ERROR   manager failed to run command   {"error": "invalid config: Key: 'System.ModelLoaders' Error:Field validation for 'ModelLoaders' failed on the 'required' tag"}
main.main
        /workspace/cmd/main.go:50
runtime.main
        /usr/local/go/src/runtime/proc.go:271
```

I was trying to play around with the branch and see how the helm validation worked.

internal/apiutils/requests.go
```diff
@@ -0,0 +1,37 @@
...
+	// Alternatives considered:
+	//
+	// "-" (hyphen): This is a common separator in Kubernetes resource names.
+	// "." (dot): This is a common separator in model versions "llama-3.2".
+	// "/" (slash): This would be incompatible with specifying model names inbetween slashes in URL paths (i.e. "/some-api/models/<model-id>/details").
+	// ":" (colon): This might cause problems when specifying model names before colons in URL paths (see example below).
+	//
+	// See example of a path used in the Gemini API (https://ai.google.dev/gemini-api/docs/text-generation?lang=rest):
+	// "https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash:generateContent?key=$GOOGLE_API_KEY"
+	adapterSeparator = "_"
```
samos123, line 18: Appreciate you sharing the alternatives as comments! I could see a future where a user needs it changed, but we can tackle it whenever that happens.

internal/modelcontroller/engine_vllm.go
```diff
@@ -27,24 +28,19 @@ func (r *ModelReconciler) vLLMPodForModel(m *kubeaiv1.Model, c ModelConfig) *cor
...
-				SecretKeyRef: &corev1.SecretKeySelector{
-					LocalObjectReference: corev1.LocalObjectReference{
-						Name: r.HuggingfaceSecretName,
-					},
-					Key:      "token",
-					Optional: ptr.To(true),
-				},
-			},
+			// https://docs.vllm.ai/en/latest/models/lora.html#dynamically-serving-lora-adapters
+			Name:  "VLLM_ALLOW_RUNTIME_LORA_UPDATING",
```
samos123, line 39: I prefer to make this conditional to only be enabled when adapters are set like you did with `--enable-lora` flag

internal/modelcontroller/adapters.go
```diff
@@ -0,0 +1,210 @@
...
+		case strings.ToLower(v1.VLLMEngine):
+			param.engine = v1.VLLMEngine
+		default:
+			continue
+		}
+
+		deletionCandidates := getLabelledAdapters(pod)
+
+		for _, adapter := range adapters {
+			if k8sutils.GetLabel(pod, v1.PodAdapterLabel(adapter.Name)) != k8sutils.StringHash(adapter.URL) {
```
samos123, line 45: Feel free to ignore this. Just writing down notes as I get through the review.

I don't get  this comparison. This should always be false if we're comparing regular string against a string hash.

This is from docs  "// StringHash returns a hash value calculated from the input string."

So shouldn't this be this instead:
```
if k8sutils.GetLabel(pod, v1.PodAdapterLabel(adapter.Name)) != adapter.URL {
```

Maybe GetLabel returns a hash as well? then it would make sense.

internal/modelcontroller/pod_utils.go
```diff
@@ -0,0 +1,65 @@
...
+			Stdout:    true,
+			Stderr:    true,
+		}, runtime.NewParameterCodec(r.Scheme))
+
+	exec, err := remotecommand.NewSPDYExecutor(r.RESTConfig, "POST", execReq.URL())
+	if err != nil {
+		return fmt.Errorf("creating remote command executor: %w", err)
+	}
+
+	if err := exec.Stream(remotecommand.StreamOptions{
```
samos123, line 33: not sure if this is important or not:

exec.Stream is deprecated: use StreamWithContext instead to avoid possible resource leaks. See https://github.com/kubernetes/kubernetes/pull/103177 for details

VS code complained about it.

internal/modelcontroller/engine_vllm.go
```diff
@@ -149,7 +145,13 @@ func (r *ModelReconciler) vLLMPodForModel(m *kubeaiv1.Model, c ModelConfig) *cor
 		},
 	}
 
+	r.patchServerAdapterLoader(&pod.Spec, m, r.ModelLoaders.Image)
```
samos123, line 148: Would prefer same logic here to only patch when adapters are found in model:
```
if m.Spec.Adapters != nil {
```

Reason being, preventing the impact if for some reason this feature breaks. We shouldn't impact all users.

samos123 (changes requested):
I didn't finish all. Leaving partial feedback, will continue more tomorrow.

nstogner:
> Not sure whether this should work or not. So sharing what I did and the error message I get:
> 
> I was trying to play around with the branch and see how the helm validation worked.

Here is what I used to make sure that the most recent image was running:

```bash
gcloud container clusters create-auto cluster-1 \
    --location=us-central1

skaffold run -f ./skaffold.yaml --tail --port-forward --profile kubeai-only-gke --default-repo us-central1-docker.pkg.dev/substratus-dev
```

internal/modelcontroller/engine_vllm.go
nstogner: That logic lives inside of `patchServerAdapterLoader`, but agree it might be clearer to put it here.

internal/modelcontroller/pod_utils.go
nstogner: That does sound like it would be good to do

api/v1/model_types.go
```diff
@@ -21,18 +21,34 @@ import (
...
-	// For VLLM & FasterWhisper engines: "hf://<model-repo>/<model-name>"
-	// For OLlama engine: "ollama://<model>
+	// Currently the following formats are supported:
+	//
+	// For VLLM, FasterWhisper, Infinity engines:
+	//
+	// "hf://<repo>/<model>"
+	// "gs://<bucket>/<path>" (only with cacheProfile)
+	// "oss://<bucket>/<path>" (only with cacheProfile)
+	// "s3://<bucket>/<path>" (only with cacheProfile)
```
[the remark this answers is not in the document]
nstogner: As implemented, the model loader is ran as a sidecar with a shared volume for adapters. The Model.spec.url is not passed to the model loader container, it is only passed to the engine (i.e. vLLM).

internal/modelcontroller/engine_vllm.go
nstogner: So this works the same as the rest of the patch functions that only patch if needed:
```go
	r.patchServerAdapterLoader(&pod.Spec, m, r.ModelLoaders.Image)
	patchServerCacheVolumes(&pod.Spec, m, c)
	c.Source.modelAuthCredentials.applyToPodSpec(&pod.Spec, 0)
```
I have it this way to avoid a bunch of conditionals in all of the engine-specific Pod funcs

internal/modelcontroller/engine_vllm.go
nstogner: Done

internal/modelcontroller/pod_utils.go
nstogner: Updated

samos123 (approved):

internal/modelcontroller/engine_vllm.go
samos123: I missed that. Makes sense to do it that way.

alpe:
nice work!

PR #307, opened 2024-11-07
Llama 3.1 70b with pipeline parallelism

samos123:


nstogner (approved):
It would be good to add comments into the Model settings so that we know why each value was set. But its also good to merge as-is.

PR #310, opened 2024-11-16
update vllm image for GPU and TPU to v0.6.4.post1

samos123:
Note the new GH200 image is not available yet and will be updated later.


samos123:
oh weird, seems the new CPU image is broken. Investigating further and will fix before merging.

samos123:
I have fixed a bug for vLLM and will try to get it fixed, but for now let's merge this for GPU and TPU @nstogner

nstogner (approved):

PR #313, opened 2024-11-23
Add Configure Text Generation Models guide

samos123:


docs/how-to/configure-text-generation-models.md
```diff
@@ -0,0 +1,160 @@
...
+
+KubeAI supports the following engines for text generation models (LLMs, VLMs, ..):
+
+- vLLM (Recommended for GPU)
+- Ollama (Recommended for CPU)
+- Need something else? Please file an issue on [GitHub](https://github.com/substratusai/kubeai).
+
+There are 2 ways to install a text generation model in KubeAI:
+- Use Helm with the `kubeai/models` chart.
+- Use `kubectl apply -f model.yaml` to install a Model CRD.
```
nstogner, line 11: `s/CRD/Custom Resource` - CRD is the definition of a CR

docs/how-to/configure-text-generation-models.md
```diff
@@ -0,0 +1,160 @@
...
+
+KubeAI comes with pre-validated and optimized Model configurations for popular text generation models. These models are available in the `kubeai/models` Helm chart and
```
nstogner, line 13: Prefer links to GitHub directories.

docs/how-to/configure-text-generation-models.md
```diff
@@ -0,0 +1,160 @@
...
+are also published as raw manifests in the `manifests/model` directory.
+
+You can also easily define your own models using the Model CRD directly or by using the `kubeai/models` Helm chart.
+
+## Install a Text Generation Model using Helm
+
+KubeAI provides a `kubeai/models` chart that contains the pre-configured models.
```
nstogner, line 20: Already state above

docs/how-to/configure-text-generation-models.md
```diff
@@ -0,0 +1,160 @@
...
+
+You can take a look at all the pre-configured models in the chart's default values file.
```
nstogner, line 22: Link would be useful

docs/how-to/configure-text-generation-models.md
```diff
@@ -0,0 +1,160 @@
...
+
+```bash
+helm show values kubeai/models
```
nstogner, line 25: They would need to have loaded the chart first right?

docs/how-to/configure-text-generation-models.md
```diff
@@ -0,0 +1,160 @@
...
+    engine: VLLM
+    resourceProfile: nvidia-gpu-l4:1
+    minReplicas: 1 # by default this is 0
+EOF
+```
+
+## Install a Text Generation Model using kubectl
+You can use the Model CRD directly to install a model using `kubectl apply -f model.yaml`.
+
+### Install Text Generation Model using CPU
```
nstogner, line 61: I dont think we need to duplicate the steps for CPU... I think we could just point to an example model after defining steps for install models above.

docs/how-to/configure-text-generation-models.md
```diff
@@ -0,0 +1,160 @@
...
+            }
+        ]
+    }'
+```
+
+### Using the OpenAI Python SDK to interact with the model
+Once the pod is ready, you can use the OpenAI Python SDK to interact with the model:
+All OpenAI SDKs work with KubeAI since the KubeAI service is OpenAI API compatible.
+
+See the below example code to interact with the model using the OpenAI Python SDK:
```
nstogner, line 138: Alt: we could just link to different examples to keep docs easier to maintain. This would have the added benefit of directing users to the repo where they could add stars, issues, etc.

nstogner (changes requested):

docs/how-to/configure-text-generation-models.md
samos123: That's a given if you're trying to configure models after you've installed KubeAI. Instlaling KubeAI requires adding the KubeAI helm repo.

docs/how-to/configure-text-generation-models.md
samos123: As a user I prefer to have the full examples in the doc. We can link to more examples in the repo but I want to at least show CPU and GPU for both helm and non-helm flow.

docs/how-to/configure-text-generation-models.md
samos123: I think we should embed code from the repo in docs in the future, but I do think from user pov that having it in the same doc is preferred.

docs/how-to/configure-text-generation-models.md
samos123: done

docs/how-to/configure-text-generation-models.md
samos123: done

docs/how-to/configure-text-generation-models.md
samos123: done

samos123:
Not sure whether this should work or not. So sharing what I did and the error message I get:
```
git checkout $THIS_PR

k get pods
NAME                         READY   STATUS             RESTARTS     AGE
kubeai-6bf98d5b77-txmpb      0/1     CrashLoopBackOff   1 (2s ago)   5s
openwebui-69ffb7dbb4-xcvlj   0/1     Running            0            5s

k logs -f kubeai-6bf98d5b77-txmpb 
2024-11-24T06:37:29Z    INFO    manager run finished
2024-11-24T06:37:29Z    ERROR   manager failed to run command   {"error": "invalid config: Key: 'System.ModelLoaders' Error:Field validation for 'ModelLoaders' failed on the 'required' tag"}
main.main
        /workspace/cmd/main.go:50
runtime.main
        /usr/local/go/src/runtime/proc.go:271
```

I was trying to play around with the branch and see how the helm validation worked.

PR #314, opened 2024-11-24
Proposal: Cache-optimized routing

nstogner:
Goal: Route requests to take advantage of prefix caching.

Proposal: Use hashing to define preferred backends for given requests. This avoids the need to keep track of the state of backend caches.

Reference: https://research.google/blog/consistent-hashing-with-bounded-loads/


samos123:
Can you explain more how the hashing works?

nstogner:
Been doing some research, will likely ditch the diagrammed approach for something like:

https://github.com/buraksezer/consistent
https://medium.com/vimeo-engineering-blog/improving-load-balancing-with-a-new-consistent-hashing-algorithm-9f1bd75709ed

PR #326, opened 2024-12-02
add nvidia-gpu-rtx4070-8gb and qwen2.5 models

samos123:
@nstogner curious what you think about this pattern of using memory requirements as profiles. Note this is for my laptop where I have an NVIDIA GPU of 8GB memory. It's an RTX 4070 and the laptop has version has 8GB whereas desktop version has 12GB.


nstogner:
I think it would be better to define two profiles for `RTX 4070`, similar to what we do with the different memory variants of a100: https://github.com/substratusai/kubeai/blob/d834b1c4206fe44b108484ad614448741510d540/charts/kubeai/values.yaml#L153

Issue #311, opened 2024-11-19
Proposal: Mount a PVC in ReadManyOnly mode as model storage


samos123:
Use case: Users have pre-provisioned PVs that contain models on them and support ReadManyOnly. The user would be responsible for ensuring a compatible model is stored on the PV and creating a PVC.

Example:
```
model.url: "pvc://$PVC_NAME/$PVC_MODEL_PATH"
```

The $PVC_MODEL_PATH will always be an absolute path startign with a `/`.

Model is stored in PV under /llama: `model.url: "pvc://123f124/llama"` or  "pvc://123f124//llama"` 

If the model is stored under the root directory in PV, then this would both be valid:
```
`model.url: "pvc://123f124/
`model.url: "pvc://123f124
```

The following would happen inside the model engine pod:
1. Mount the PVC named $PVC_NAME on the $PV_PATH to engine container on path `/model`
2. The engine would always load the model using `/model`

Open questions:
1. Should we expose the mounting mode ReadManyOnly vs ReadWriteOnce?
    No this won't be needed since the user is responsible for creating the PVC. So the user will have control over this. The only issue is we would start encountering issues when we try to scale beyond 1 replica with ReadWriteOnce.

Example user flow:
1. User creates a PVC using ReadManyOnly and binds this PVC to an existing PV that has the model loaded under `/llama-3-8b`. The PVC name is `llama-3-8b`
2. User creates a `Model` and specifies model.url as `pvc://llama-3-8b/llama-3-8b`
3. KubeAI: The model engine gets configured to mount the PVC

Step 1 PVC:
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: llama-3-8b
  labels:
    type: local
spec:
  storageClassName: robin
  resources:
    requests:
      storage: 10Gi
  accessModes:
    - ReadOnlyMany
  volumeName: name-of-pv-with-prepulated-llama-3-8gb
```

Step 3 pod spec:
```
        volumeMounts:
        - mountPath: "/model"
          subPath: llama-3-8b
          name: model-pvc
    volumes:
      - name: model-pvc
        persistentVolumeClaim:
          claimName: llama-3-8b
          readOnly: true
```

## Why not have KubeAI manage the PVC and allow users to specify PV instead?
It allows the user to specify more attributes that may be relevant on the PVC. One example that wouldn't be as easy is figuring out the resource.requests.storage capacity to request for the PVC. So it may make more sense to have the user control that.

[...]

PR #339, opened 2024-12-05
vLLM: Add support for loading models from PVC

### LINKED ISSUE #303: how to access model files by pvc?

samos123:
Implements #311 
Fixes #303

Validation logic is implemented to prevent a Model with PVC source and cacheProfile from being admitted.

```
k apply -f - << EOF
apiVersion: kubeai.org/v1
kind: Model
metadata:
  name: opt-125m-cpu-pvc
  namespace: default
spec:
  engine: VLLM
  features:
  - TextGeneration
  minReplicas: 2
  replicas: 2
  resourceProfile: cpu:1
  scaleDownDelaySeconds: 30
  targetRequests: 100
  url: pvc://model-pvc
  cacheProfile: test
EOF
The Model "opt-125m-cpu-pvc" is invalid: spec: Invalid value: "object": cacheProfile is only supported with urls of format "hf://...", "s3://...", "gs://...", or "oss://..." at the moment.
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,12 @@ jobs:
     #needs: unit-and-integration # No use in running e2e tests if integration tests fail.
     strategy:
       matrix:
-        testcase: ["quickstart", "openai-python-client", "autoscaler-restart", "cache-shared-filesystem"]
+        testcase:
+        - "quickstart"
+        - "openai-python-client"
+        - "autoscaler-restart"
+        - "cache-shared-filesystem"
+        - "engine-vllm-pvc"
     steps:
       - name: Checkout code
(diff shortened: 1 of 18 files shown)


nstogner (approved):
Looks good. One suggestion, add a test case for the cache profile validation: https://github.com/substratusai/kubeai/blob/a12fecfad85384ab107b2a959d0e01dc025c57a9/test/integration/model_validation_test.go#L249

PR #360, opened 2025-01-01
Benchmark prefix hashing on 8 replicas using H100 and Llama 3.1 70B

samos123:


charts/models/values.yaml
```diff
@@ -139,6 +139,18 @@ catalog:
       - --disable-log-requests
     resourceProfile: nvidia-gpu-h100:2
     targetRequests: 500
+  llama-3.1-70b-instruct-fp8-1xh100:
```
nstogner, line 142: For the rest of the models we dont specify "1x<gpu>" any reason to do so here?

nstogner (commented):
Added a question, other than that, LGTM

charts/models/values.yaml
samos123: because there already is a llama-3.1-70b-instruct-fp8 model that uses  2 x h100 and has bigger context length. I don't recommend changing model names of existing models. So suggest adding this new model using the 1xh100 naming convention.

PR #379, opened 2025-02-02
feat: switch to upstream openwebui helm chart

### LINKED ISSUE #285: Look into decommissioning embedded Open WebUI chart in favor of the upstream chart
See: https://github.com/open-webui/helm-charts

kaiehrhardt:
fixes #285


samos123:
Thanks Kai! This is very helpful. Initially when we started there was no good openwebui helm chart, but now there is.

We may eventually switch to no longer including openwebui by default but let's do that in a separate PR.

Let me know if you need help with getting tests to pass again.

charts/kubeai/values.yaml
```diff
@@ -209,38 +209,27 @@ messaging:
   streams: []
 
 # Configure the openwebui subchart.
-openwebui:
-  fullnameOverride: "openwebui"
-  service:
-    type: ClusterIP  # ClusterIP, NodePort, LoadBalancer
-    port: 80
```
samos123, line 216: Note to self. It's fine to remove this since this is the default: https://github.com/open-webui/helm-charts/blob/921af388489fac825a722fad2d26469b8ad94ff5/charts/open-webui/values.yaml#L197

samos123:
Do we need to include this file in the repo itself? charts/kubeai/charts/open-webui-5.4.0.tgz

I was assuming specify the dependency chart in Chart and chart repo url was enough.

charts/kubeai/Chart.yaml
```diff
@@ -18,9 +18,10 @@ appVersion: "v0.14.0"
 dependencies:
   # Open Web UI is an open source ChatGPT-like user interface.
   # https://docs.openwebui.com/
-  - name: openwebui
-    condition: openwebui.enabled
-    version: 0.2.0
+  - name: open-webui
+    condition: open-webui.enabled
+    repository: https://helm.openwebui.com/
```
samos123, line 23: why do we need to include the tgz file itself if we already specify the repo here? I probably prefer to remove the tgz file and just point to upstream helm repo url.

charts/kubeai/values.yaml
```diff
@@ -209,38 +209,27 @@ messaging:
...
-    # TODO: This changes with .fullnameOverride and .service.port, make this more robust.
-    value: "http://kubeai/openai/v1"
-  # A good number of features are not compatible with the KubeAI architecture.
   - name: ENABLE_OLLAMA_API
     value: "false"
   - name: SHOW_ADMIN_DETAILS
     value: "false"
   - name: SAFE_MODE
     value: "true"
-  - name: ENABLE_LITELLM
```
samos123, line 239: do we no longer need to set this explicitly?

kaiehrhardt: It seems like, there is no `ENABLE_LITELLM` anymore, check https://docs.openwebui.com/getting-started/env-configuration/.

charts/kubeai/Chart.yaml
kaiehrhardt: I have asked myself the same question. Does it still work without the tgz? I will try that.

kaiehrhardt: Without the tgz:
```
Error: INSTALLATION FAILED: An error occurred while checking for chart dependencies. You may need to run `helm dependency build` to fetch missing dependencies: found in Chart.yaml, but missing in charts/ directory: open-webui
```

samos123: Will it automatically download the tgz file somehow and we can simply put it in .gitignore?

samos123: I will play around with this a bit more as well to see if we can figure out a way to omit these tgz files. If not, we can just include them as well.

skaffold.yaml
```diff
@@ -24,6 +24,9 @@ deploy:
     releases:
     - name: kubeai
       chartPath: ./charts/kubeai
+      setValueTemplates:
+        open-webui.enabled: false
+      skipBuildDependencies: true
```
kaiehrhardt, line 29: i'm not sure about that. do you want openwebui in your dev env?

charts/kubeai/Chart.yaml
samos123: Generated Artifacts:
The .tgz file created by helm dependency build is a generated artifact. Since it’s derived from the dependency information in your Chart.yaml, it can be recreated at any time by running that command.

Version Control Best Practices:
It’s generally best practice to avoid committing generated files to Git. Instead, you should commit the source (in this case, the dependency definitions in Chart.yaml) and let the build process handle generating the required artifacts.

Vendoring vs. Rebuilding:
The only time you might consider committing the .tgz files is if you need your chart to be completely self-contained (for example, to support offline installations or to ensure absolute reproducibility without fetching dependencies). However, if your users can run helm dependency build as part of their workflow, it’s simpler and cleaner not to commit them.

Publishing as a Helm Repo:
When you package your chart for publication in a Helm repo (using helm package), the packaging process will include your dependencies appropriately. The repository consumers will download the packaged chart, not the raw Git repo, so committing the .tgz file is unnecessary.

^^ AI had this to say. I think we're probably fine with not including the tgz as long as we run `helm dependency build` during local development. The publishing will automatically include the depedencies. I will make some changes to this PR to try this out.

skaffold.yaml
samos123: We can leave as-is for now. We probably should have 1 test that includes it at a minimum though.

samos123:
I am seeing the following warning when testing locally:
```
W0205 21:53:19.307473 1133781 warnings.go:70] spec.template.spec.containers[0].env[4]: hides previous definition of "ENABLE_OLLAMA_API", which may be dropped when using apply
NAME: kubeai
LAST DEPLOYED: Wed Feb  5 21:53:18 2025
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

Will try to fix.

samos123:
This new openwebui has some Arena Chat model that we should try to remove imo
image

samos123:
Confirmed it's all working well locally. Didn't know about the preview feature in Open WebUI

image

samos123 (approved):

charts/kubeai/Chart.yaml
samos123: I was able to remove the tgz file from repo and it should be included in the release packaging workflow as well. I will make sure to validate that in the upcoming release.

samos123:
Thanks for your patience @kaiehrhardt I made a few follow up commits. Could you please review my changes before I merge it?

The main things I changed are:
* Remove .tgz file from git
* Update our testing and releasing workflows
* Remove Arena
* Fix warning

kaiehrhardt:
lgtm

PR #390, opened 2025-02-10
Benchmarking prefix cache routing

samos123:


nstogner:
These results look pretty promising. A few things:

1. Since we have so much a gap in TTFT between PrefixHash and LeastLoad load balancing modes, I think it would make a lot of sense to add in a 3rd scenario to sanity check: raw k8s Service (i.e. round-robin). This will give us an indication of whether there is a bug in the LeastLoad codepath that might be causing the ramp in TTFT. This can be done by setting `minReplicas: 8` in the Model, and applying a new Service that will bypass KubeAI (set `--base-url=http://bypass-kubeai` in the benchmark script):

```yaml
apiVersion: v1
kind: Service
metadata:
  name: bypass-kubeai
spec:
  selector:
    model: llama-3.1-8b-instruct-fp8-l4
  ports:
    - protocol: TCP
      port: 80
      targetPort: 8000
```

2. I am wondering why TTFT actually gets shorter at medium load for PrefixHash routing - any thoughts? Perhaps it is just within the variance that you might expect across different runs.

3. The graph is nice, would love to see one for the throughput numbers as well.

4. I am assuming you are keeping the Model object around across the different test runs and just changing the load balancing strategy field? Or are you deleting and recreating?

samos123:
1. Will do another test.
2. I think that's just pure variance between runs.
3. Will do.
4. Yes I only modified the model by patching the load balancing strategy. This caused the pods to stay alive.

benchmarks/benchmark_serving/benchmark_serving.py
```diff
@@ -0,0 +1,1373 @@
+# SPDX-License-Identifier: Apache-2.0
```
nstogner, line 1: Can you add comment with a link to where this file came from and what was changed?

docs/README.md
```diff
@@ -19,6 +20,24 @@ Quotes from the community:
 
 > reusable, well abstracted solution to run LLMs - [Mike Ensor](https://www.linkedin.com/posts/mikeensor_gcp-solutions-public-retail-edge-available-cluster-traits-activity-7237515920259104769-vBs9?utm_source=share&utm_medium=member_desktop)
 
+## Why KubeAI?
+
+### Better performance at scale
```
nstogner, line 25: +Intro: `When running multiple replicas of a serving engine such as vLLM, performance under production traffic is heavily influence by the load balancing strategy.`

benchmarks/benchmark_serving/README.md
```diff
@@ -0,0 +1,16 @@
+# Benchmarking Text Generation
+
+This script was adopted from the vLLM code base. The main differences are:
+- Load the whole conversation as prompts.
+- Limit the amount of max conversations and re-use the same conversation if needed.
+
+This allows us to verify whether prefix aware load balancing provides a performance
```
nstogner, line 7: `s/at large scale/under heavy production traffic with ongoing chat conversations/`

benchmarks/benchmark_serving/README.md
```diff
@@ -0,0 +1,16 @@
+# Benchmarking Text Generation
```
nstogner, line 1: We use `kebab-case` in all of the other dirs in this project. Prefer to keep with that convention and rename this to `benchmarks/chat-py/` and change the other to be `benchmarks/chat-k6/`.

nstogner (changes requested):
Great asset to have this benchmark!

docs/README.md
samos123: I will leave this as is but please do a follow up PR to modify the README.

benchmarks/benchmark_serving/README.md
samos123: done

benchmarks/benchmark_serving/benchmark_serving.py
samos123: done

benchmarks/benchmark_serving/README.md
samos123: done

PR #409, opened 2025-02-14
Add deepseek R1 1.5b ollama model

samos123:


charts/models/values.yaml
```diff
@@ -408,6 +408,12 @@ catalog:
       - --enforce-eager
     engine: VLLM
     resourceProfile: 'nvidia-gpu-l4:1'
+  deepseek-r1-1.5b-ollama:
```
nstogner, line 411: Current convention is to suffix with `-cpu`. I would prefer to keep with that, or refactor all model names.

nstogner (changes requested):
Good switch, asking to change one name, other than that, feel free to merge

charts/models/values.yaml
samos123: done

nstogner (approved):

PR #414, opened 2025-02-16
Multi-turn chat benchmark implemented in Go

nstogner:
Create a new benchmark in Go to capture metrics from a multi-turn chat scenario that were not possible with the k6 benchmark (i.e. TTFT).

During the process of benchmarking, a major issue was found the the `PrefixHash` load balancing strategy: when Pods were terminated mid-load, the in-flight requests would be double-decremented from the `totalInFlight` measurement - throwing the "bounded load" check off for future requests. This issue was fixed in this PR.


benchmarks/multi-turn-chat-go/example/pod.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: bench
+spec:
+  restartPolicy: Never
+  containers:
+    - name: bench
+      image: us-central1-docker.pkg.dev/substratus-dev/default/benchmark-multi-turn-chat-go
+      args: ["--config=/config/config.json", "--threads=./data/tiny.json"]
```
samos123, line 10: I personally prefer passing the config as individual CLI args. Removes the need to configure volume, volumeMount and configmaps.

nstogner: I'll add that as an option

Makefile
```diff
@@ -131,7 +131,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
 docker-build: ## Build docker image with the manager.
-	$(CONTAINER_TOOL) build -t ${IMG} .
+	$(CONTAINER_TOOL) build --platform=linux/amd64 -t ${IMG} .
```
samos123, line 134: Not sure if this would break anything? I guess we don't rely on Makefile in our build and docker image release process so it's fine.

samos123: for example, we publish arm images today.

internal/loadbalancer/balance_chwbl.go
```diff
@@ -15,14 +17,23 @@ func (g *group) chwblGetAddr(key string, loadFactor float64, adapter string) (en
 	}
 
 	h := chwblHash(key)
-	_, i0 := g.chwblSearch(h)
+	hash0, i0 := g.chwblSearch(h)
+	//log.Printf("hash: %v, i0: %v", h, i0)
```
samos123, line 21: remove this

samos123 (approved):

internal/loadbalancer/balance_chwbl.go
nstogner: done

Makefile
nstogner: Whoops, I should revert this

PR #417, opened 2025-02-19
chore:  use podLister with informer instead of client.List for better performance

Fleezesd:
Replace the direct `client.List` call with `PodLister.List` in the Model controller to improve pod listing performance.

## Motivation
- Using informer cache instead of direct API server calls
- Better performance for pod listing operations
- Following Kubernetes controller best practices
- Reducing API server load

## Changes
- Using informerFactory to make informer cache
- Replaced `client.List` with `PodLister.List` for pod listing
- Using label selector with `labels.SelectorFromSet`

## Testing
- [✅] Verified informer cache syncs correctly
- [✅] Confirmed pod listing functionality works as expected
- [✅] Tested cache update mechanism

### DIFF
--- a/internal/manager/run.go
+++ b/internal/manager/run.go
@@ -14,6 +14,7 @@ import (
 	"sigs.k8s.io/yaml"
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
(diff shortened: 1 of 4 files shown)


nstogner:
I will take a look at this tomorrow! Thanks for the PR.

nstogner:
`Using informer cache instead of direct API server calls` --> the existing `Client` from the controller runtime package already uses an informer-based cache that is wired up by the controller `Manager`.

Fleezesd:
> `Using informer cache instead of direct API server calls` --> the existing `Client` from the controller runtime package already uses an informer-based cache that is wired up by the controller `Manager`.

okay, I re-read the controller-runtime source code, and  update my own project structure for use，thanks for you review code ! 👀

PR #429, opened 2025-03-01
Make url mutable when cacheProfile is not used

### LINKED ISSUE #422: Allow Updating Model URL for Seamless Model Replacement
### Description:

Currently, the model resource in KubeAI has an immutable url field, preventing updates to the model’s source URL. This design necessitates deleting and recreating the entire model object to change the model, which disrupts continuous service and complicates rolling updates.

### Problem:

- Immutability of url Field: The inability to modify the url field restricts flexibility in updating models.
- Service Disruption: Deleting and recreating the model object leads to downtime, as there is no seamless transition between the old and new models.
- Lack of Rolling Updates: The current approach doesn’t support rolling updates, making it challenging to update models without affecting ongoing services.

samos123:
Partially implements #422 

Making the url mutable when cacheProfile is used will be done in a follow up PR.

Also updates the quickstart e2e test to use deepseek to correctly reflect what happens in the quickstart.


test/e2e/quickstart/test.sh
```diff
@@ -20,4 +24,36 @@ EOF
...
+
+# Test to ensure that model url can be updated without requests failing
+kubectl patch model deepseek-r1-1.5b-cpu --type=merge -p '{"spec": {"url": "ollama://qwen2.5:0.5b"}}'
+
+# Set maximum number of retries
+MAX_RETRIES=120
+retry_count=0
+
+# Continiously run curl requests to the model until the new pod is ready
+while true; do
```
nstogner, line 40: Prefer to use retry func:

https://github.com/substratusai/kubeai/blob/bc6c60fd50535c2802a1b274dbd175e5d637b355/test/e2e/common.sh#L3

nstogner (changes requested):
I think the e2e test should be updated to provide some more certainty that the rollout worked successfully. I think it would be good to list the pods for the model `-l model=...` running at the end and make sure that the container args contains the new model url.

test/e2e/quickstart/test.sh
nstogner: ps: it is already included since this script is imported

liaddrori1:
Thanks for the quick turnaround!

I have one question—why is it mutable only when cacheProfile is not used?

Currently, we’re downloading models directly from Hugging Face, but our DevOps team is working on adding EFS to our clusters to cache the models for faster spin-up times. Ideally, we’d like to cache the models while still maintaining the ability to update them.

Again, I really appreciate your help!

nstogner:
The plan is to support both. Breaking it out into 2 PRs (the latter one implementing cache updates) allows us to get the feature out quicker.

nstogner:
@samos123 we should also include documentation on what to expect during the rollout process and what to expect if the new Pods fail to become ready during the rollout.

liaddrori1:
This is great. 
Again - appreciate the quick help!

samos123:
> @samos123 we should also include documentation on what to expect during the rollout process and what to expect if the new Pods fail to become ready during the rollout.

Let's do that in a separate PR since that may require more back and forth reviews.

test/e2e/quickstart/test.sh
```diff
@@ -6,9 +6,13 @@ models_release="kubeai-models"
```
nstogner, line 6: The test for this functionality should be in its own e2e test, not the quickstart test... This rollout is not conducted in the quickstart.

nstogner (changes requested):

nstogner:
> > @samos123 we should also include documentation on what to expect during the rollout process and what to expect if the new Pods fail to become ready during the rollout.
> 
> Let's do that in a separate PR since that may require more back and forth reviews.

I think it would worth doing here in order to fully understand the impact of this change. I have forgotten the details myself.

test/e2e/quickstart/test.sh
samos123: I think it makes sense to extend tests. The e2e test for quickstart is also testing the quickstart steps.

There is a limit on how many parallel tests we can execute. Quickstart made the most sense to me. In a follow up PR I'm extending the test for cacheprofile to also test mutating the URL.

samos123:
Created a follow up issue to tackle docs: https://github.com/substratusai/kubeai/issues/435 Can you add anything else you want documented there?

The reason I prefer to handle that separately is because I don't want to delay this PR.

test/e2e/quickstart/test.sh
samos123: I basically don't want to replicate the quickstart in another test just so I can test mutating the url field.

nstogner (approved):
I would prefer to use lowercase var names in bash for non-env script variables, but I am good to merge as-is.

PR #437, opened 2025-03-03
Add imagePullSecrets Support for Model Pods

### LINKED ISSUE #433: Feature Request: Support for Image Pull Secrets in Model CRD to Address Docker Hub Rate Limits
###  Background
Docker Hub has announced that starting April 1, 2025, unauthenticated users will be limited to only 10 pulls per hour.
Source: [Docker Hub Usage and Limits](https://docs.docker.com/docker-hub/usage/#:~:text=Starting%20April%201%2C%202025%2C%20all,Unauthenticated%20users%3A%2010%20pulls%2Fhour)

### Issue
Currently, the Model CRD does not support specifying image pull secrets directly. This limitation makes it difficult to use authenticated Docker Hub pulls or private images with KubeAI models.

### Feature Request
Add support for image pull secrets in the Model CRD to allow users to specify credentials for pulling container images.

### Proposed Solution
Add an imagePullSecrets field to the Model CRD spec, similar to how it works in Pod specs:
```yaml
apiVersion: kubeai.org/v1
kind: Model
metadata:
  name: my-model
spec:
  # ... existing fields ...
  imagePullSecrets:
  - name: my-docker-credentials
```

### Edit - Another Proposed Solution by @samos123:
Setting the image pull secrets array at the KubeAI config level 

### Benefits

- Helps users avoid Docker Hub rate limits by authenticating pulls
- Allows for using private images in model deployments
- Improves user experience in environments with multiple models

### Additional Context
This feature becomes increasingly important as Docker Hub enforces stricter rate limits. Without it, users may experience failed pulls and service disruptions when deploying models.

liaddrori1:
# Add imagePullSecrets Support for Model Pods
## Description

This PR adds support for propagating imagePullSecrets from the main KubeAI configuration to model pods. This enables model pods to pull images from private registries using the same credentials as the main server.

## Changes

- Added `ImagePullSecrets` field to `ModelServerPods` configuration struct
- Updated chart values.yaml to document imagePullSecrets
-  Updated Helm chart configmap template to pass imagePullSecrets from values to system config
- Updated model engine reconcilers to pass ImagePullSecrets to pod specifications
- Added integration tests to verify ImagePullSecrets propagation

## Example Usage

In your Helm values.yaml:
```yaml
imagePullSecrets:
- name: my-registry-secret
- name: another-registry-secret
```

The secrets will automatically be used by all model pods created by KubeAI.

## Related Issues
Closes #433

### DIFF
--- a/charts/kubeai/templates/configmap.yaml
+++ b/charts/kubeai/templates/configmap.yaml
@@ -31,6 +31,10 @@ data:
       securityContext:
         {{- .Values.modelServerPods.securityContext | toYaml | nindent 8}}
       {{- end}}
+      {{- if .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml .Values.imagePullSecrets | nindent 8}}
+      {{- end}}
       {{- end}}
       serviceAccountName: {{ include "models.serviceAccountName" . }}
     modelAutoscaling:
--- a/charts/kubeai/values.yaml
+++ b/charts/kubeai/values.yaml
@@ -252,6 +252,7 @@ image:
   # Overrides the image tag whose default is the chart appVersion.
   tag: ""
 
+# The imagePullSecrets will be used to pull both the kubeai controller image and the model images.
 imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: "kubeai"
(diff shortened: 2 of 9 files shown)


nstogner:
@liaddrori1 Looks great. The only question in my head: did you run through a local helm install using this change? Asking b/c we are missing a test case that asserts on end-to-end plumbing of this being set in helm values all the way through to the model Pods (which I think is OK, just wanted to hear if this was manually tested as well).

liaddrori1:
I tested it locally. 
The setup was: 
Changes to skaffold.yaml
image

setting up a model and checking the created pod
image


Let me know if anything else is needed.

nstogner:
Fantastic, thanks for the contribution!

PR #440, opened 2025-03-04
add goatcounter analytics to kubeai.org website

samos123:
Privacy-aware; doesn’t track users with unique identifiers and doesn't need a GDPR notice.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -4,6 +4,7 @@ repo_url: https://github.com/substratusai/kubeai
 
 theme:
   name: material
+  custom_dir: docs/overrides
   palette:
     primary: white # Defaults to indigo.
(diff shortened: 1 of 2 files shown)


mkdocs.yml
```diff
@@ -54,3 +55,8 @@ markdown_extensions:
   - pymdownx.tasklist:
       custom_checkbox: true
   - pymdownx.tilde
+
+# Analytics tracking with GoatCounter
```
nstogner, line 59: What is this block for?

samos123: to specify overrides so we can add analytics to docs website.

samos123: https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-analytics/#custom-site-analytics

PR #449, opened 2025-03-06
Strongly typed OpenAI payloads

nstogner:
* Move `api/v1` package to `api/k8s/v1`
* Create `api/openai/v1` package
* Introduce types for OpenAI API
  * Started with https://github.com/sashabaranov/go-openai & extensively updated to maintain lossless json-struct-json roundtrip semantic meaning
   * NOTE: Only the *Request types are actually used by the proxy at this point - until we expand to support response modification


api/openai/v1/chat.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	openai "github.com/sashabaranov/go-openai"
+)
+
+type ChatCompletionRequest struct {
+	openai.ChatCompletionRequest
+}
+
+func (r *ChatCompletionRequest) GetModel() string {
```
samos123, line 11: How would we implement something like a Model router with this approach? https://github.com/lm-sys/RouteLLM

nstogner: Will follow up on this in a later PR

samos123 (approved):

PR #461, opened 2025-03-17
E2e test updates

nstogner:
* Use self-contained skaffold files for e2e tests (remove reliance on one large skaffold file with multiple profiles)
* Remove redundancy in make `test-e2e-*`


samos123 (approved):
Looks good, please merge. Follow up, I prefer to enable open web ui for e2e tests so we can catch any issues that may arise there too. Since it's enabled by default in our helm charts.

PR #470, opened 2025-03-24
fix: Include minReplicas even when value is 0

27Bslash6:
## Issue
The current Helm template for models omits the `minReplicas` field when its value is set to 0. This happens because the Go template uses a `with` statement:

~~~yaml
{{- with $model.minReplicas }}
minReplicas: {{ . }}
{{- end }}
~~~

In Go templates, the integer value `0` is treated as "empty", causing the field to be completely absent from the generated Kubernetes resources.

## Impact
When `maxReplicas` is defined but `minReplicas` is missing, the Kubernetes validation fails with:

~~~
Error: failed to create resource: Model.kubeai.org is invalid: spec: Invalid value: "object": no such key: minReplicas evaluating rule: minReplicas should be less than or equal to maxReplicas.
~~~

This prevents models from being deployed with scaling configurations that use `minReplicas: 0` (scale-to-zero), which is a common pattern for efficient resource usage.

## Solution
Replace the conditional inclusion with a default value approach to ensure the field is always present:

~~~yaml
minReplicas: {{ default 0 $model.minReplicas }}
~~~

This guarantees that the field is included in the YAML even when the value is 0, satisfying the Kubernetes validation rule.

## Testing
Tested by deploying models with `minReplicas: 0` configurations, confirming they now pass validation and deploy successfully.

This example fails on the current version, but succeeds with the PR code:

```yaml
# CPU
deepseek-r1-1.5b-cpu:
  enabled: true
  features: [TextGeneration]
  url: "hf://deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B"
  engine: VLLM
  minReplicas: 0
  maxReplicas: 1
```

### DIFF
--- a/charts/models/templates/models.yaml
+++ b/charts/models/templates/models.yaml
@@ -30,9 +30,7 @@ spec:
   env:
   {{- toYaml . | nindent 4 }}
   {{- end }}
-  {{- with $model.minReplicas }}
-  minReplicas: {{ . }}
-  {{- end }}
+  minReplicas: {{ default 0 $model.minReplicas }}
   {{- with $model.maxReplicas }}
   maxReplicas: {{ . }}
   {{- end}}
--- a/manifests/models/bge-embed-text-cpu.yaml
+++ b/manifests/models/bge-embed-text-cpu.yaml
@@ -7,4 +7,5 @@ spec:
   features: [TextEmbedding]
   url: hf://BAAI/bge-small-en-v1.5
   engine: Infinity
+  minReplicas: 0
   resourceProfile: cpu:1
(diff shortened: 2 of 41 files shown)


samos123:
Thanks for making a PR and raising this issue! Looks good to me overall.

Can you also run `make manifests`, commit and push?

samos123 (changes requested):
See previous comment.

27Bslash6:
Done!

samos123 (approved):

samos123:
I will do a new release of KubeAI soon that will include this PR. Trying to get vLLM images updated before doing a new release.

PR #478, opened 2025-03-26
fix: use correct accessKeyID values in if statement

ffais:
Sorry but in the previous PR I forgot to update the if directive with the correct accessKeyID value. 

I also updated the comments in the values ​​and the documentation with the new value.


samos123 (approved):

samos123:
Nothing to be sorry about! Really appreciate you catching this and doing the follow up. We don't have active e2e tests for s3.

Issue #451, opened 2025-03-08
LoRA Adapters from  S3 Compatible Storage Server


samos123:
### Discussed in https://github.com/substratusai/kubeai/discussions/450

<div type='discussions-op-text'>

<sup>Originally posted by **tuckkin** March  7, 2025</sup>
Hi.. may I know if there is any timeline for LoRA adapters to be loaded from PVC (PV with NFS) and also the S3 Compatible Storage Server such as MinIO? Thank you.
https://www.kubeai.org/how-to/serve-lora-adapters/</div>

PR #482, opened 2025-03-30
Load models and LoRA Adapters from S3 Compatible Storage Server

ffais:
Support load models and LoRA Adapters from non-AWS S3 Compatible Storage Server allowing users to specify the endpoint URL.

I've also added an e2e test to verify that loading models from s3 works. However I couldn't find a way to easily test the loading of adapters from s3.

Fixes: #451

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,6 +37,7 @@ jobs:
         - "engine-vllm-pvc"
         - "engine-ollama-pvc"
         - "model-files"
+        - "s3-model"
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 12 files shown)


ffais:
Hi @nstogner I made the changes you requested, let me know what you think.

poussa:
Cool. This would also solve the issue of running the model loader behind corporate firewall. That is, if you define `https_proxy` in `Model.Spec.Env`, that should endup to the model loader container and at lease huggingface-cli should work. Not sure about the other sources.

ffais:
It should works for S3 too, since AWS cli uses HTTPS_PROXY if it is present.

samos123 (approved):

samos123:
Thank you for this great PR. LGTM but let's wait for @nstogner to review too since he did an initial review already.

nstogner:
Great job on the PR!

PR #491, opened 2025-04-04
feat: Allow Ollama insecure pull using ?insecure=true in url

### LINKED ISSUE #488: OLlama: Allow pulling from "insecure" registry
I would like to use a local cache registry to provide models to Ollama.

Ollama supports the `--insecure` flag for both pushing and pulling images from a registry that does not run with HTTPS. I am in a network where it would be an unnecessary overhead to create a valid HTTPS certificate to host a local-only registry.

In the end, what I would like for KubeAI to do is when I provide a model URL like `ollama://hostname:port/namespace/model-name:model-tag` that I can also specify it is to be pulled with the insecure flag. So basically run `ollama pull --insecure hostname:port/namespace/model-name:model-tag` instead of just `ollama pull hostname:port/namespace/model-name:model-tag` as it does now.

samos123:
Fixes #488

### DIFF
--- a/docs/how-to/configure-text-generation-models.md
+++ b/docs/how-to/configure-text-generation-models.md
@@ -148,15 +148,16 @@ chat_completion = client.chat.completions.create(
 )
 ```
 
-## Ollama Insecure Model Pulling
+## Ollama Configuration Notes
 
-You can set the environment variable `INSECURE=true` to allow pulling from insecure repositories.
+### Insecure Model Pulling
+
+To allow pulling Ollama models from insecure registries (e.g., HTTP-only or self-signed TLS), append the `?insecure=true` query parameter to the model URL.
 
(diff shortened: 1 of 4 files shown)


nstogner:
I would prefer to avoid using an env var just as a means of getting the info to the right place where we can then set a flag. I think a better approach is to pass this info along in the model url. We already established the convention of using query parameters in our urls to support loading ollama models from PVCs ([see example](https://github.com/substratusai/kubeai/blob/995ded2585190005dc492729dcfa741c11ec2a08/internal/modelcontroller/model_source_test.go#L118)). I would prefer to build on that convention here: `url: ollama://qwen2?insecure=true`.

nstogner (changes requested):
Added a comment above ^

nstogner (approved):
LGTM

PR #493, opened 2025-04-05
fix flaky s3 test take 2

samos123:


samos123:
I re-ran the flaky test 3 times and all succeeded. Hopefully it's not just luck this time.

PR #497, opened 2025-04-07
support use of JSONPatch for model server pods

alam0rt:
Adds support for applying arbitrary JSON 6902 / JSONPatch support for model server pods.

For example, our clusters have a number of mutating webhooks which inject special priority class names and various pre-emption policies based on various labels, etc. This gives operators a way to set the pod templates up which can work in constrained clusters (i.e. those with OPA gatekeeper constraints).

```yaml
modelServerPods:
  - op: add
    path: /spec/priorityClassName
    value: gpu

```

### DIFF
--- a/charts/kubeai/templates/configmap.yaml
+++ b/charts/kubeai/templates/configmap.yaml
@@ -27,6 +27,10 @@ data:
       podSecurityContext:
         {{- .Values.modelServerPods.podSecurityContext | toYaml | nindent 8}}
       {{- end}}
+      {{- if .Values.modelServerPods.podPatches }}
+      podPatches:
+        {{- .Values.modelServerPods.podPatches | toYaml | nindent 8}}
+      {{- end}}
       {{- if .Values.modelServerPods.securityContext }}
       securityContext:
         {{- .Values.modelServerPods.securityContext | toYaml | nindent 8}}
--- a/charts/kubeai/templates/configmap.yaml
+++ b/charts/kubeai/templates/configmap.yaml
@@ -27,9 +27,9 @@ data:
       podSecurityContext:
         {{- .Values.modelServerPods.podSecurityContext | toYaml | nindent 8}}
       {{- end}}
-      {{- if .Values.modelServerPods.podPatches }}
-      podPatches:
-        {{- .Values.modelServerPods.podPatches | toYaml | nindent 8}}
+      {{- if .Values.modelServerPods.jsonPatches }}
+      jsonPatches:
+        {{- .Values.modelServerPods.jsonPatches | toYaml | nindent 8}}
       {{- end}}
       {{- if .Values.modelServerPods.securityContext }}
       securityContext:
--- a/internal/config/system.go
+++ b/internal/config/system.go
@@ -233,7 +233,7 @@ type ModelLoading struct {
 	Image string `json:"image" validate:"required"`
 }
 
-type Patch struct {
+type JSONPatch struct {
 	Op    string      `json:"op"`
 	Path  string      `json:"path"`
 	Value interface{} `json:"value"`
@@ -252,8 +252,8 @@ type ModelServerPods struct {
 	// ImagePullSecrets is a list of references to secrets in the same namespace to use for pulling any of the images
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
 
-	// ModelPodPatches is a list of patches to apply to the model pod template.
+	// JSONPatches is a list of patches to apply to the model pod template.
 	// This is a JSON Patch as defined in RFC 6902.
 	// https://datatracker.ietf.org/doc/html/rfc6902
-	ModelPodPatches []Patch `json:"podPatches,omitempty"`
+	JSONPatches []JSONPatch `json:"jsonPatches,omitempty"`
 }
--- a/internal/config/system.go
+++ b/internal/config/system.go
@@ -233,6 +233,12 @@ type ModelLoading struct {
 	Image string `json:"image" validate:"required"`
 }
 
+type Patch struct {
+	Op    string      `json:"op"`
+	Path  string      `json:"path"`
+	Value interface{} `json:"value"`
+}
+
 type ModelServerPods struct {
 	// The service account to use for all model pods
 	ModelServiceAccountName string `json:"serviceAccountName,omitempty"`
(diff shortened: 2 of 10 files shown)


nstogner:
Thanks for the PR! Will try to get back to you on this one by tomorrow.

alam0rt:
> Thanks for the PR! Will try to get back to you on this one by tomorrow.

cheers mate! I have been using in our clusters and it does what it says on the box. Was originally going to just add a way to specify priority class / preemption policy but figured arbitrary patches would make for a cleaner change.

internal/config/system.go
```diff
@@ -245,4 +251,9 @@ type ModelServerPods struct {
 
 	// ImagePullSecrets is a list of references to secrets in the same namespace to use for pulling any of the images
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+
+	// ModelPodPatches is a list of patches to apply to the model pod template.
+	// This is a JSON Patch as defined in RFC 6902.
+	// https://datatracker.ietf.org/doc/html/rfc6902
+	ModelPodPatches []Patch `json:"podPatches,omitempty"`
```
nstogner, line 258: Rename to `JSONPatches` (json: `jsonPatches`). This is within the context of `ModelServerPods` ... so we don't need the duplicate "Pod" part of the name. The JSON part specifies what kind of patch this is.

internal/modelcontroller/patch.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/substratusai/kubeai/internal/config"
+	jsonpatch "gopkg.in/evanphx/json-patch.v4"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func patchPod(patches []config.Patch, pod *corev1.Pod) error {
```
nstogner, line 12: Rename to `applyJSONPatchToPod()`

internal/modelcontroller/patch.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return fmt.Errorf("marshal pod patch: %w", err)
+	}
+
+	patch, err := jsonpatch.DecodePatch(pb)
+	if err != nil {
+		return fmt.Errorf("decode pod patch: %w", err)
+	}
+
+	podB, err := json.Marshal(pod)
```
nstogner, line 27: Rename `podB` to something like `podJSON`.

internal/modelcontroller/patch.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return fmt.Errorf("marshal pod: %w", err)
+	}
+
+	patchedPodB, err := patch.Apply(podB)
```
nstogner, line 32: `patchedPodJSON`

internal/modelcontroller/patch_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/substratusai/kubeai/internal/config"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+)
+
+func Test_patchPod(t *testing.T) {
+	t.Parallel()
+
+	cases := map[string]struct {
```
nstogner, line 16: Add a case for applying `len = 0` patches.

internal/modelcontroller/pod_plan.go
```diff
@@ -37,6 +37,8 @@ func (r *ModelReconciler) calculatePodPlan(allPods *corev1.PodList, model *kubea
 	default:
 		podForModel = r.vLLMPodForModel(model, modelConfig)
 	}
+
+	patchPod(r.ModelServerPods.ModelPodPatches, podForModel)
```
nstogner, line 41: Check error

charts/kubeai/templates/configmap.yaml
```diff
@@ -27,6 +27,10 @@ data:
       podSecurityContext:
         {{- .Values.modelServerPods.podSecurityContext | toYaml | nindent 8}}
       {{- end}}
+      {{- if .Values.modelServerPods.podPatches }}
+      podPatches:
```
nstogner, line 31: Rename to `jsonPatches`

nstogner (changes requested):
Great contribution! We would like to merge this. Requested some changes.

internal/modelcontroller/pod_plan.go
alam0rt: Have added an error check, since `calculatePodPlan()` doesn't return an error of its own, I've opted to feed in the `context.Context` from the `Reconcile()` loop in order to log that there was a failure applying the patches and that they will be ignored. Plus added a test in `pod_plan_test.go` to ensure that an invalid patch will not change the expected pod hash.

internal/modelcontroller/model_controller.go
```diff
@@ -173,7 +173,7 @@ func (r *ModelReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 		}
 	}()
 
-	plan := r.calculatePodPlan(allPods, model, modelConfig)
+	plan := r.calculatePodPlan(ctx, allPods, model, modelConfig)
```
alam0rt, line 176: fed this in for the purpose of logging any error from applying the patches.

internal/modelcontroller/patch_test.go
```diff
@@ -0,0 +1,117 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+)
+
+func Test_applyJSONPatchToPod(t *testing.T) {
+	t.Parallel()
+
+	emptyPod := &corev1.Pod{}
+	emptyPodHash := k8sutils.PodHash(emptyPod.Spec)
+
```
alam0rt, line 19: Not sure if this is overkill, but thought I'd reuse the pod hash util to have a better guarantee that the pod has not been modified.

nstogner:
I'll take another look tomorrow, thanks for updating

nstogner:
I would like to update the error handling logic to follow the general pattern that exists in the project: return an error as far up the stack as possible until it can be handled or logged at the highest point possible. Instead of logging in the `calculatePodPlan()` func, let's update that func to return an error. Inside of `Reconcile()`, if that error is returned, log the error, then return a nil error - because returning a non-nil error in Reconcile will start a retry-loop - and retrying is not going to solve the issue until the Model is updated - in which case Reconcile() will be called again naturally.

We will add a followup PR to fire a k8s event from within Reconcile() to bubble the error up to the admin on the Model object - we have several other use cases for events, so I would worry about that part in this PR.

nstogner (changes requested):
Looks good, just requesting we change error handling pattern, see above comment

alam0rt:
May need to, in a subsequent PR, think of ways to target specific engines/model server pods. i.e. if we want to inject patches into _just_ VLLM.

nstogner (approved):
LGTM!

PR #504, opened 2025-04-17
allow s3 model urls without cache profile

liebman:
There are a few small issues that prevent loading llms via s3 without using a cache profile:
- remove rule enforcing s3 only with cache profile
- if no cache profile and scheme is s3 then use original url and not a pvc

### DIFF
--- a/api/k8s/v1/model_types.go
+++ b/api/k8s/v1/model_types.go
@@ -24,7 +24,6 @@ import (
 
 // ModelSpec defines the desired state of Model.
 // +kubebuilder:validation:XValidation:rule="!has(self.cacheProfile) || self.url.startsWith(\"hf://\") || self.url.startsWith(\"s3://\") || self.url.startsWith(\"gs://\") || self.url.startsWith(\"oss://\")", message="cacheProfile is only supported with urls of format \"hf://...\", \"s3://...\", \"gs://...\", or \"oss://...\" at the moment."
-// +kubebuilder:validation:XValidation:rule="!self.url.startsWith(\"s3://\") || has(self.cacheProfile)", message="urls of format \"s3://...\" only supported when using a cacheProfile"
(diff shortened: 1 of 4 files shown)


samos123:
This is great! Thanks a bunch for testing and fixing what's needed to get it to work.

I pushed changes from `make manifests` to this PR so the new CRD is updated with the validation rule.

Could you share the full Model spec that worked with loading a model from S3?

liebman:
This is what I used:

```yaml
apiVersion: kubeai.org/v1
kind: Model
metadata:
  annotations: {}
  labels:
    features.kubeai.org/TextGeneration: 'true'
  name: meta-llama-3.1-8b-instruct-fp8
  namespace: llm-infra
spec:
  args:
  - --load-format=runai_streamer
  - --max-model-len=16384
  - --max-num-batched-token=16384
  - --gpu-memory-utilization=0.9
  - --disable-log-requests
  engine: VLLM
  env:
    AWS_EC2_METADATA_DISABLED: "true"
    RUNAI_STREAMER_S3_USE_VIRTUAL_ADDRESSING: "0"
    RUNAI_STREAMER_NO_BOTO3_SESSION: "1"
    AWS_ENDPOINT_URL: "http://s3-compat-storage-url-goes-here:8080"
  features:
  - TextGeneration
  loadBalancing:
    prefixHash:
      meanLoadFactor: 125
      prefixCharLength: 100
      replication: 256
    strategy: LeastLoad
  maxReplicas: 1
  minReplicas: 0
  owner: k8s
  replicas: 1
  resourceProfile: nvidia-gpu-l4:1
  scaleDownDelaySeconds: 2400
  targetRequests: 50
  url: "s3://llm-models/Meta-Llama-3.1-8B-Instruct-FP8"

```

samos123:
One test is failing: --- FAIL: TestModelValidation/s3-url-without-cache-profile-invalid (0.00s)

I am removing the test.

samos123:
It's available in KubeAI v0.20.0 that was just released @liebman

PR #506, opened 2025-04-21
Add priorityClassName to the model spec and pass to generated pods

liebman:
Enable use of  Kubernetes [Pod Priority and Preemption](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/) to configure your models with different priority levels. This is useful when you have limited resources and want to ensure that high-priority models can preempt lower-priority models when necessary.


liebman:
Just rebased on latest main.

samos123 (approved):

samos123:
Thank you! PR looks great to me. Will give @nstogner a chance to review too.

nstogner:
I would prefer to avoid exposing Pod spec options on a field-by-field basis as we will trend towards replicating the entire Pod spec in the Model spec. Instead, I would prefer to support patching via the Model spec. Similar to what we do at the system config-level: https://github.com/substratusai/kubeai/blob/d6e393ca76f11da76b3a6db74b737b94d1a4f057/internal/config/system.go#L258

liebman:
This won't work either as an int or string won't be accepted as an `object` right?
```
	// Value is the value to be used in the operation.
	// Required for "add", "replace", and "test" operations.
	// +kubebuilder:validation:Optional
	// +kubebuilder:validation:Schemaless
	// +kubebuilder:validation:Type=object
	Value interface{} `json:"value,omitempty"`
```

samos123:
Thanks for giving it a quick try. Let's keep this PR open because I do see value in exposing priorityClassName in a more straight forward way with defaulting to it not being set.

liebman:
@samos123 Any chance this can be merged? (We'd like to deploy this to production soon) It Looks like the simplest way to support setting priority.  I also created a PR to support JSONPatch (#516) its limited in that the value field of a patch must be a string containing the JSON encoded value. (A simple string value has to be specified as "\"simple-string\"")

samos123:
Thanks for the follow up! It's been busy recently. Apologies for the delay.

@nstogner I think this is a clean small addition for a somewhat common use case. The JSONPatch PR is a bit more tricky and clunky for just needing to set priortyClass.

nstogner:
LGTM

samos123:
Thanks for your patience @liebman ! Hopefully this unblocks you for now.

PR #516, opened 2025-04-28
feat: support JSONPatch in ModelSpec

liebman:
This PR adds JSONPatch support to the Model CRD.

There is difficulty representing the `Value` field in JSONPatch, its a go `interface{}` but there seems to be no type to map it to in CRDs. So this is implemented as a string expected to contain the JSON encoded value,  For instance if the value is a string: `"\"this-is-a string-value\""`

### DIFF
--- a/test/integration/model_jsonpatch_test.go
+++ b/test/integration/model_jsonpatch_test.go
@@ -4,22 +4,23 @@ import (
 	"testing"
 	"time"
 
+	v1 "github.com/kubeai-project/kubeai/api/k8s/v1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	v1 "github.com/substratusai/kubeai/api/k8s/v1"
 	corev1 "k8s.io/api/core/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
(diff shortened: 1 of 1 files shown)


liebman:
I added a test that tries to set a string value and as expected when defining the Type as object it fails 
```
    model_jsonpatch_test.go:40: 
                Error Trace:    kubeai/test/integration/model_jsonpatch_test.go:40
                Error:          Received unexpected error:
                                Model.kubeai.org "testjsonpatch" is invalid: [spec.jsonPatches[0].value: Invalid value: "string": spec.jsonPatches[0].value in body must be of type object: "string", <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]
                Test:           TestJSONPatch
```

If I don't set a Type it errors loading the CRD complaining there is no type.

liebman:
The only way I can think of to make this work is to store Value as a string containing json.  So a string value would be encoded as "\"string-value\"", an integer as "123" and an object as "{\"something\": 42}" and I don't like it as it's not what the user would expect, especially in a Model yaml file, the value would still need to be a encoded as json in a string.

liebman:
Implemented where `JSONPatch.value` is a string expected to contain valid JSON.

PR #518, opened 2025-04-30
Support multiple Kubeai installation in different namespace

ffais:
At the moment it is not possible to install multiple Kubeai instances in the same cluster and in different namespaces due to the way helm handles CRDs.

In this PR I added the _crds.enabled_ value to control the installation of CRDs.

An alternative would be to move the "crds" folder to the root of the Chart and use the helm flag --skip-crds to disable its installation. However, this involves having to manually manage the CRDs update.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,12 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen yq
 	# Generate CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) crd webhook paths="./..." output:crd:artifacts:config=charts/kubeai/templates/crds/
+	$(CONTROLLER_GEN) crd webhook paths="./..." output:crd:artifacts:config=manifests/crds/
+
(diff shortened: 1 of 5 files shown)


charts/kubeai/Chart.yaml
```diff
@@ -7,7 +7,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.20.0
+version: 0.20.1
```
samos123, line 10: Can you revert this? I handle this as part of release process.

samos123 (changes requested):
Lgtm except change of helm chart version.

charts/kubeai/Chart.yaml
ffais: Done!

ffais:
@samos123 @nstogner any news on this?

PR #528, opened 2025-05-19
update chwbl load balancing to remove +1 to load

mskouba:
Update chwbl load balancing to not add one to the endpoint's load when determining whether an endpoint can accommodate  an additional request. Per the conversation in #519, this solution helps handle traffic better in situations where there are many endpoints and many requests with a common prefix.


samos123:
Appreciate your patience and taking the time to come up with a small clean improvement

Issue #542, opened 2025-06-18
Ollama engine in disconnected (airgapped) environments


buroa:
There is a problem with running images in airgapped environments.

My container image has a local ollama image, but the startup probe pulls the manifest even though its present. This will not work when the environment is disconnected.

PR #543, opened 2025-06-18
fix(ollama): only pull manifest image if not present

buroa:
this will fix deployments that reside in airgapped environments

fixes https://github.com/substratusai/kubeai/issues/542


buroa:
@samos123 I updated this. I also added a new `schedulerName` field to the pod specs and updated the unit tests.

samos123:
Ideally each PR adds 1 feature / bug fix only. That way when there is an issue with 1 of the things we can easily revert them independently.

Lgtm otherwise. Fine to merge these 2 together.

PR #669, opened 2026-06-15
Feat/dra resource profile

DavideRutigliano:
Extends #642 (original work by @futomaru).

Rebased onto current main; no functional changes.

## Summary
- Adds `spec.resourceClaimName` to the Model CR for per-model DRA support
- Injects `resourceClaims` into model server Pods and `resources.claims` into containers, using a fixed claim name `gpu-claim` / request `gpu`
- Updates CRD manifests, Helm chart, values, and API reference docs

## Notes
- Claim name and request are fixed to `gpu-claim` / `gpu` (MVP)
- Do not combine `modelServerPods.jsonPatches` with `resourceClaimName` on the same model — the global patch applies to all pods

## Testing
- `make generate`
- `make manifests`
- `make fmt`
- `make vet`
- `make test-unit`

### DIFF
--- a/charts/kubeai/values.yaml
+++ b/charts/kubeai/values.yaml
@@ -217,6 +217,32 @@ resourceProfiles:
         operator: "Equal"
         value: "present"
         effect: "NoSchedule"
+  # DRA profile examples - Dynamic Resource Allocation (Kubernetes 1.32+).
+  # DRA profiles are mutually exclusive with GPU entries in limits.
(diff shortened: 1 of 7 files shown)


ffais:
I was thinking of putting the DRA configuration inside the resource profile definition of kubeai instead of inside the model definition.
What do you guys think?

DavideRutigliano:
> DRA configuration inside the resource profile definition

@ffais agree, that was the whole idea behind extending @futomaru PR

PR #691, opened 2026-07-12
proxy: preserve original request body when no rewrite is needed

### LINKED ISSUE #690: Proxy re-serializes request bodies with non-deterministic JSON key order, breaking prompt caching in all backends
The proxy seems to re-serializes every JSON request body, even when it doesn't change anything: `readJSONBody` in `internal/apiutils/request.go` unmarshals into the `openaiv1` structs and on my setup always replaces the body with `json.Marshal(...)`.

Untyped fields like `FunctionDefinition.Parameters any` become `map[string]any` on the way, and go-json-experiment marshals maps in non-deterministic order. So two byte-identical requests leave the proxy with the tool schemas shuffled differently each time.

That silently kills prompt/prefix caching in every backend behind kubeai (vLLM, Ollama): the rendered prompt differs from the first tool definition onwards on every request, so nothing past that point can ever be reused. In our agentic setup (~22k token sessions, coding agent with 29 tools) every single turn was a full re-prefill, 90+ seconds instead of ~1s.

Took us a while to find because every component looked innocent in isolation. After implementing a llama compatibility layer to kubeai, llama-server's slots debug finally showed it on accident:

```
old: ...  "parameters": {" | type": "object", "
new: ...  "parameters": {" | required": ["path"], "
```

The client verifiably sent byte-identical bodies (we captured them in front of the proxy), and replaying a captured body twice directly against the backend gave a perfect cache hit. Only the proxy hop broke it.

Suggested fix: keep the original body bytes when there's nothing to rewrite (adapter is empty), and only re-marshal when the model field actually has to change. Alternatively `json.Deterministic(true)` would at least make the order stable. We're running the first variant as a local patch and it fixes caching completely. Happy to send a PR.

krim404:
Fixes #690.

The proxy currently re-marshals every request body even when it doesn't change anything. Untyped fields like tool parameter schemas round-trip through maps, and jsonv2 writes those in non-deterministic key order - so byte-identical client requests reach the backend with their tool schemas shuffled differently every time. Any prompt/prefix cache in the backend (vLLM, llama.cpp, Ollama) stops matching after the first tool definition.

This keeps the original body bytes unless the model field actually has to be rewritten for an adapter.

We run this in production: an identical 22k-token agentic request went from a full re-prefill (90+ s on our hardware) to a ~1 s cache hit. Existing apiutils tests pass unchanged.

### DIFF
--- a/internal/apiutils/request.go
+++ b/internal/apiutils/request.go
@@ -180,7 +180,12 @@ func (r *Request) readJSONBody(body io.Reader, path string) error {
 		return fmt.Errorf("unknown path: %q", path)
 	}
 
-	if err := json.UnmarshalRead(body, r.modelRequest); err != nil {
+	raw, err := io.ReadAll(body)
+	if err != nil {
+		return fmt.Errorf("reading: %w", err)
+	}
+
+	if err := json.Unmarshal(raw, r.modelRequest); err != nil {
 		return fmt.Errorf("decoding: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


DavideRutigliano:
Thanks for the fix and great catch. The PR also looks good.

Before merging, could you add a regression test so we don't quietly fall into this behavior later? I think it would fit well into `internal/apiutils/request_test.go`.

Something like:

1. Add a new case covering the exact issue you had (a tool schema with non-alphabetically ordered keys that would trigger reshuffle during serialization)
```
{
	name: "chat completion with tool schema",
	body: `...`,
	path: "/v1/chat/completions",
	expModel:  "test-model",
	expPrefix: "test",
}
```

2. Add an assertion in in `t.Run` loop, after the existing `require.Equal` calls:

```
if c.expAdapter == "" {
	require.Equal(t, c.body, string(req.Body),
	"no-adapter path must forward the client body byte-for-byte")
}
```

krim404:
Sure no problem, i'll push it in a few minutes.

krim404:
finished. sorry for the force push, accidentally pushed the example instead of the finished test.
```
│ PR-Branch  │ 20× │ PASS (20/20) │
│ main │ 20× │ FAIL (20/20) │
```

DavideRutigliano (approved):

ffais (approved):

PR #733, opened 2026-09-15
fix: execute Ollama preparation before startup probe succeeds

### LINKED ISSUE #721: Startup probe regression: bash -c array form silently no-ops for pvc:// Ollama models (v0.23.4)
## Bug

Since PR #656 (v0.23.2), the Ollama engine startup probe for `pvc://` URL models is broken. The `ollama cp` alias and `ollama run` preload commands never execute — the probe passes immediately without doing anything useful.

## Root cause

PR #656 fixed command injection by switching to an array exec form, but **left `/bin/bash -c` in the array and spread the remaining command words as separate elements**:

```json
["/bin/bash", "-c", "/bin/ollama", "cp", "deepseek-r1:1.5b", "deepseek-r1-1.5b",
 "&&", "/bin/ollama", "run", "deepseek-r1-1.5b", "hi"]
```

`bash -c SCRIPT` takes the **next single argument** as the script string. Every element after `-c` becomes a positional parameter (`$0`, `$1`, …), not part of the command. So bash runs `/bin/ollama` with zero arguments (exits 0), and `cp` / `run` / `&&` are silently ignored.

## Effect

- `ollama cp` never runs → the KubeAI-internal model name alias is missing → inference returns `404 model not found`
- `ollama run <model> hi` never runs → model weights aren't preloaded into VRAM → first request cold-starts

The startup probe passes immediately (exit 0 from bare `ollama`), so the pod goes Ready with a non-functional model backend.

## Fix

Join the command words into a single string after `-c`:

```go
// Before (broken):
[]string{"/bin/bash", "-c", "/bin/ollama", "cp", src, dst, "&&", "/bin/ollama", "run", dst, "hi"}

// After (correct):
[]string{"/bin/bash", "-c", fmt.Sprintf("/bin/ollama cp %s %s && /bin/ollama run %s hi",
    shellescape(src), shellescape(dst), shellescape(dst))}
```

Note: since inputs are already validated/sanitized by the security fix in #656, interpolating them into a single `-c` string is safe as long as the sanitization regex remains in place.

## Reproduction

Deploy any `pvc://` Ollama model, then inspect the startup probe on the running pod:

```bash
kubectl get pod <model-pod> -n <ns> \
  -o jsonpath='{.spec.containers[0].startupProbe.exec.command}'
# Output: ["/bin/bash","-c","/bin/ollama","cp",...]
#                              ^^^^^^^^^^^^^
#                 This is $0, not part of the script!

# The KubeAI alias is never created:
kubectl exec -n <ns> <model-pod> -- ollama list
# Source tag present, <kubeai-model-name> alias absent

# Inference 404:
curl http://kubeai/openai/v1/chat/completions \
  -d '{"model":"deepseek-r1-1.5b","messages":[...]}'
# {"error":"model not found"}
```

## Version

`v0.23.4` confirmed against live cluster. Regression introduced in v0.23.2 via #656.

## Workaround

After every pod start, manually create the alias:

```bash
kubectl exec -n <ns> <pod> -- ollama cp <source-tag> <kubeai-model-name>
# e.g.:
kubectl exec -n gpu model-deepseek-r1-1.5b-xxxxx -- ollama cp deepseek-r1:1.5b deepseek-r1-1.5b
```

ruslan-shaydullin:
## Problem

The Ollama startup probe passes command words separately after `bash -c`, so Bash invokes `/bin/ollama` without arguments and can report success without pulling, aliasing or preloading the model. This affects PVC and registry URLs reported in #721 by kadomino and juancarlosm.

## Change

- Pass one fixed script with quoted positional model arguments and explicit CLI option terminators. Stop at the first failed pull/copy/preload step.
- Preserve #656 URL validation, #543 `pull=false`, PVC handling, insecure pulls, embedding-only models and existing probe timings.
- Execute the generated Pod probe in regression tests, checking argv, order, repeated execution and each failure point. Add parser/CRD negative cases.
- Strengthen the CPU PVC e2e to check alias, preload and successful completion on each replica, then the routed API; align the hydrate image with Ollama 0.11.11.

Fixes #721.

## Validation

On macOS arm64 with Go 1.25.1 (`GOMAXPROCS=3 GOFLAGS=-p=3`):

- `go test -v ./internal/modelcontroller -run 'TestOllamaStartupProbeExec|Test.*ModelURL' -count=1`: passes. The same tests on the unchanged base fail all 36 probe cases with one empty invocation. Only `/bin/ollama` is substituted by a temporary argv logger; these are native shell tests.
- `make test-unit`: passes, including fmt/vet.
- `make test-integration`: passes with envtest Kubernetes 1.33.0.
- `bash -n test/e2e/engine-ollama-pvc/test.sh` and `git diff --check`: pass.
- `make lint`: blocked building pinned golangci-lint v1.59.1 (`x/tools@v0.22.0/internal/tokeninternal`: invalid array length `-256`); the same failure occurs on the base; lint analysis did not run.

Actual CPU runtime validation also passes on an isolated arm64 Kind 0.31.0 / Kubernetes 1.35.0 cluster using the manager built from this PR and Ollama 0.11.11 (the chart default):

- `make test-e2e TEST_CASE=engine-ollama-pvc`: both replicas have the alias loaded according to `/api/ps` before their first test completion, then return HTTP 200 with the requested model and nonblank text; the routed API also succeeds.
- Additional local checks on two recreated PVC pods execute each stored startup probe again, with unchanged alias/source digests, pod UIDs and container restart counts, followed by successful direct completions.
- A separate local harness checks `ollama://registry.ollama.ai/library/qwen:0.5b` at initial startup, after deleting the pod, and after scaling to zero. Initial and replacement pods have alias/preload before the first completion; direct and routed requests succeed. After zero pods are observed, a routed request creates a third pod and succeeds; its UID is captured before increasing `minReplicas` and retained afterward.
- With the real Ollama CLI, `pull --help` succeeds as help; `pull -- --help`, `cp -- --help safe-alias` and `run -- --help hi` each fail as invalid model references.

These checks establish functional CPU serving, not model answer quality. Ollama 0.32.15 from the issue and GPU execution were not tested. The test cluster has been removed.

Draft pending the pinned-linter/toolchain issue above and upstream workflows, which currently require maintainer approval (`action_required`).

AI assistance: implementation and an independent code review used Codex.


ruslan-shaydullin:
Could a maintainer approve the three pending GitHub Actions workflows on this PR? No workflow checks have executed yet. The PR remains Draft while upstream validation is outstanding; local unit/integration and Ollama PVC runtime checks are recorded in the description, along with the pinned lint-toolchain compatibility blocker. Thanks!

